PR #14, opened 2021-04-06
feat(php): use prebuild binaries

### LINKED ISSUE #9: Use new php-prebuild

viceice:
Use our prebuild php binaries.

closes #9


viceice:
Should we backport this to `renovatebot/docker-*` images ?

rarkins:
> Should we backport this to `renovatebot/docker-*` images ?

Will it take long? If so then maybe next time the ubuntu base image changes?

viceice:
I think that would be fast when it's working. There seem to be some extension issues which I need to investigate and fix first

rarkins (approved):

viceice:
image


needs another fix for bionic 😕 Will do that later

PR #15, opened 2021-04-07
feat: add Git LFS

### LINKED ISSUE #13: Add git-lfs
* download linux-amd64 from [releases](https://github.com/git-lfs/git-lfs/releases/latest)
* extract
* copy binary to path `/usr/local/bin` (alternate is using the included `install.sh`)
* run `git lfs install` as `ubuntu` user

blocks: https://github.com/renovatebot/renovate/issues/6842

We should use [other](https://github.com/git-lfs/git-lfs/wiki/Installation#other) install method, as its faster (install) that the apt version. We should also use this because the available version on ubuntu is old.

----
Ref https://github.com/renovatebot/docker-buildpack/issues/139

fgblomqvist:
This adds Git LFS as a tool.
I'm not too sure what kinds of tests you'd expect/want? Or where to add the `install-tool` line. It seems like by default, only git is installed in this base image?

Finally, worth noting is the discussion in https://github.com/renovatebot/renovate/issues/7894.

Closes #13.


viceice (changes requested):
Please checkout other tools like `dotnet` for download and extract tools

https://github.com/containerbase/buildpack/blob/4ca3c834bb5e96ecd0625d450a833365ffaaf686/src/usr/local/buildpack/tools/dotnet.sh#L1-L44

It should be installed to base image after git here:
https://github.com/containerbase/buildpack/blob/8717a2dd66724f98ed8b35f21cc52151885669a3/Dockerfile#L30-L31

and here:
https://github.com/containerbase/buildpack/blob/c0b1b2e6f1b3dc190f841777ce1102c9f3f919c2/Dockerfile.bionic#L27-L28

rarkins:
@viceice I was thinking we wouldn't want it in the base image (that all images share) and instead only in Renovate?

viceice:
> @viceice I was thinking we wouldn't want it in the base image (that all images share) and instead only in Renovate?

Ok, so it should be installed to `Dockerfile.bionic` and `test/latest/Dockerfile`  for testing.
We need to prepare a simple git test repo with one lfs file and unse it in tests to verify it works for `root` and `user`

fgblomqvist:
Found this: https://github.com/git-lfs/git-lfs/wiki/Installation#debian-and-ubuntu
It's essentially possible to install git-lfs and get the latest version from apt if we want. Up to you if you still prefer the tarball for speed.

viceice:
Please stay at tarball, so we have control over installed version

fgblomqvist:
Made some changes to the install script, to more-so mimmick the dotnet one you linked @viceice.
The questions that remain are:
* How do we wanna test this?
* What do we do about `git lfs install`? Right now, I install git-lfs by running the install script, which automatically runs `git lfs install` as well. However, it's user-specific which means that it only gets installed/setup for root. But yeah, not sure how we wanna go about it (e.g. what if some people using Renovate don't want git lfs?).

fgblomqvist:
Bump @viceice @rarkins.

rarkins:
If git lfs is installed by default for the container, does that cause problems for those who don't want it or need it? I had expected not, but perhaps not aware of the edge cases.

viceice:
i think it's save to install lfs, because lfs hooks are disabled by renovate renovatebot/renovate#9673

But it think we should only have this installed to main renovate image to not have any side effects on sidecar container starting to pull huge lfs objects

viceice (changes requested):
Please add a test to https://github.com/containerbase/buildpack/blob/main/test/latest/Dockerfile

src/usr/local/buildpack/tools/git-lfs.sh
```diff
@@ -0,0 +1,26 @@
...
+  echo "Invalid version: ${TOOL_VERSION}"
+  exit 1
+fi
+
+if ! [ "$(command -v git-lfs)" ]; then
+  echo "Skipping, already installed"
+  exit 0
+fi
+
+LFS_FILE="git-lfs-linux-amd64-${TOOL_VERSION}.tar.gz"
```
viceice, line 18: Make `ARCH` configurable like in 

https://github.com/containerbase/buildpack/blob/52800f94727f582808d8bc4247f380076554b769/src/usr/local/buildpack/tools/php.sh#L29
https://github.com/containerbase/buildpack/blob/52800f94727f582808d8bc4247f380076554b769/src/usr/local/buildpack/tools/php.sh#L56

src/usr/local/buildpack/tools/git-lfs.sh
```diff
@@ -0,0 +1,26 @@
...
+wget "https://github.com/git-lfs/git-lfs/releases/download/${TOOL_VERSION}/${LFS_FILE}"
```
viceice, line 19: All other tools are usiing `curl`, why are you using `wget` ?

jtbandes:
Hi, what's the status of this work? Seems like this is a necessary step to make progress on https://github.com/renovatebot/renovate/issues/6842.

viceice:
currently blocked by renovatebot/renovate#10748

viceice:
Superseeded by #104

PR #17, opened 2021-04-07
feat(composer): allow dynamic install

viceice:
Allow to run `install-tool composer 2.0.0` as user

closes renovatebot/docker-php#7

ref: renovatebot/renovate#2355. renovatebot/renovate#8497

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ LABEL maintainer="Rhys Arkins <rhys@arkins.net>" \
   org.opencontainers.image.source="https://github.com/containerbase/buildpack"
 
 #  autoloading buildpack env
-ENV BASH_ENV=/usr/local/etc/env
+ENV BASH_ENV=/usr/local/etc/env PATH=/home/$USER_NAME/bin:$PATH
 SHELL ["/bin/bash" , "-c"]
 
 ENTRYPOINT ["docker-entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ LABEL maintainer="Rhys Arkins <rhys@arkins.net>" \
   org.opencontainers.image.source="https://github.com/containerbase/buildpack"
 
 #  autoloading buildpack env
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -16,7 +16,7 @@ LABEL maintainer="Rhys Arkins <rhys@arkins.net>" \
   org.opencontainers.image.source="https://github.com/containerbase/buildpack"
 
 #  autoloading buildpack env
-ENV BASH_ENV=/usr/local/etc/env
+ENV BASH_ENV=/usr/local/etc/env PATH=/home/$USER_NAME/bin:$PATH
```
rarkins, line 19: We ideally want to avoid using `/home/$USER_NAME` in order to achieve OpenShift compliance but should we refactor that in a different PR?

viceice: This is the only option to have a fixed path for env without require loading our env file. This should also be compatible with openshift, as the directory will exist, even when openshift will run with a random user.

Our `/home/user/` directory is readable by `root` group

rarkins: Any reason it can't be within `/usr/local` in future?

viceice: Ok, this would need to make `/usr/local/bin` writable by `root` group, so we can add shell wrapper at runtime.

Maybe allowing use writing to `/usr/local` can be again a showstopper for company usage because of security?

rarkins: Is writing to `/usr/local` unsafe?

viceice: Not sure, I'm not an expert in those things. I know, that only `root` user is allowed to write there by default.

rarkins: I think you're right, in general it doesn't seem recommended to change /usr/local/bin permissions, e.g. https://unix.stackexchange.com/questions/37724/permissions-ownership-of-usr-local-bin

rarkins: I think OpenShift respects `$HOME` if configured, so we can maybe keep it as `/home/user` regardless of there being a random UID

viceice: yes, we need to make sure all required folders are writable by root group then.

rarkins (approved):

PR #28, opened 2021-04-09
refactor: use portable shebangs

### LINKED ISSUE #27: Use portable shebangs
All of our tool install scripts (except `nix` :wink:) use the `#!/bin/bash` shebang, but these aren't portable, and aren't the best practice:tm:. Changing these to `#!/usr/bin/env bash` is a simple fix.

JamieMagee:
Fixes #27


Chumper:
@rarkins @viceice Am I right that according to #27 we want to stick with `#!/bin/bash`?
If so we can close this, correct?

viceice:
I think yes.

PR #30, opened 2021-04-10
fix(python): add libpq-dev

rarkins:
Adds `libpq` when `python` is installed, as per https://github.com/renovatebot/renovate/issues/9485#issuecomment-817132534

### DIFF
--- a/src/usr/local/buildpack/tools/python.sh
+++ b/src/usr/local/buildpack/tools/python.sh
@@ -44,6 +44,7 @@ if [[ -z "${tool_path}" ]]; then
       libbz2-dev \
       libffi-dev \
       liblzma-dev \
+      libpq-dev \
       libreadline-dev \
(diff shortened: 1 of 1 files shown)


viceice (commented):
This packages are not installed when prebuild python is used.

rarkins:
What do you recommend then? I was thinking to use it for python only unless / until we heard otherwise

viceice:
This need some refactoring, so it will always install that package if it is missing an we are installing as root.

rarkins:
@viceice what are your thoughts on adding this to our python images? More context here: https://github.com/renovatebot/renovate/discussions/14596

viceice:
ok, we can do that. but not 100% sure it's enough.

maybe some bindings are not compiled on our python, so we should also add a test case here.

viceice:
doesn't work that way

PR #47, opened 2021-05-13
arm64 support

hef:


.github/workflows/build.yaml
```diff
@@ -32,6 +32,8 @@ jobs:
       TAG: ${{ matrix.distro }}
 
     steps:
+      - uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v1
```
viceice, line 36: Use our internal tools action

https://github.com/containerbase/buildpack/blob/0d1a61e374bbda2b552d708de1766be82700c56e/.github/workflows/build.yaml#L84-L87

.github/workflows/build.yaml
```diff
@@ -44,6 +46,7 @@ jobs:
 
     strategy:
       matrix:
+        platform: [linux/amd64, linux/arm64]
```
viceice, line 49: Do not use matrix. Build both platforms on same time.

.github/workflows/build.yaml
```diff
@@ -63,6 +66,8 @@ jobs:
           ]
 
     steps:
+      - uses: docker/setup-qemu-action@v1
+      - uses: docker/setup-buildx-action@v1
```
viceice, line 70: See other comment.

viceice (changes requested):

viceice:
Do not use force push or rebase, we will squash merge anyways. 😉

.github/workflows/build.yaml
hef: just to make sure I'm not missing anything:

It looks like internal-tools sets up the required environment to create arm64 builds, but I don't see a way to use it.
in the docker build function, there is no way that I can see to pass --platform=foo into buildx, even though the buildx environment looks like it can support cross building arm64.

Am I missing something?

viceice:
closing in favor of
- #713

PR #50, opened 2021-05-17
feat: support apt proxy

viceice:
Allow passing `APT_HTTP_PROXY` to configure a temporary apt proxy while building buildpack based images.

CI is using a local proxy to speedup parallel and multiple downloads.
Maybe we can use github cache to speedup more in a later pr.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -66,6 +66,9 @@ jobs:
             latest,
           ]
 
+    env:
+      TAG: ${{ matrix.tag }}
+
     steps:
       - uses: actions/checkout@v2.3.4
 
@@ -79,7 +82,7 @@ jobs:
         run: docker image ls
 
       - name: test
-        run: docker buildx build --progress plain ./test/${{ matrix.tag }}
+        run: docker buildx bake --progress plain build_test
 
   release:
     needs: [lang, distro]
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -34,6 +34,12 @@ jobs:
     steps:
       - uses: actions/checkout@v2.3.4
 
+      - name: init env
+        run: echo "APT_HTTP_PROXY=http://$(hostname -f):3142" >> $GITHUB_ENV
+
+      - name: start apt-cacher
+        run: docker buildx build --progress plain -t apt-cacher -f Dockerfile.apt-cacher && docker run -d -p 3142:3142 apt-cacher
+
       - name: test distro
         run: docker buildx bake --progress plain test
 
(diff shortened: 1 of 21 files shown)


rarkins:
Will this also work as a build arg for end users who have a permanent apt proxy and want to point the buildpack build to it?

.github/workflows/build.yaml
```diff
@@ -34,6 +35,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2.3.4
 
+      - name: start apt-cacher
+        run: docker buildx build --progress plain -t apt-cacher -f Dockerfile.apt-cacher . && docker run -d -p 3142:3142 apt-cacher
```
rarkins, line 39: Will the cache be reused across all jobs or only within-jobs?

Dockerfile.apt-cacher
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:focal@sha256:cf31af331f38d1d7158470e095b132acd126a7180a54f263d386da88eb681d93
+
+VOLUME ["/var/cache/apt-cacher-ng"]
+RUN    apt-get update && apt-get install -y apt-cacher-ng
+
+EXPOSE 3142
+CMD    chmod 777 /var/cache/apt-cacher-ng && /etc/init.d/apt-cacher-ng start && tail -f /var/log/apt-cacher-ng/*
+
+RUN cat /etc/hosts
```
rarkins, line 9: Strange that nothing "off the shelf" already existed? Seems like a useful feature

rarkins (commented):
Add documentation for end users?

Dockerfile.apt-cacher
viceice: yes, no official image, don't want to use unknown image, we can move this to a new repo if you like the general approach.

We can also install and run apt-cacher on ubuntu worker, so no need for an docker image

.github/workflows/build.yaml
viceice: currently only for the current job. can be extended later to be more persistent

Dockerfile.apt-cacher
viceice: now using native ubuntu package

README.md
```diff
@@ -7,6 +7,11 @@
 
 This repository is the source for the Docker Hub image `containerbase/buildpack`. Commits to `main` branch are automatically built and published.
 
+## Apt proxy
+
+You can pass a custom temporary Apt proxy at build or runtime when installing new packages via `APT_HTTP_PROXY` arg.
```
rarkins, line 12: Do we support any runtime apt use? Assuming the user doesn't run as root

rarkins (approved):

README.md
viceice: Only if the user is root. 😉

PR #56, opened 2021-05-18
feat: auto home handling

viceice:
- we now override `HOME=/home/$USER_NAME` if a random user is used which gets `HOME=/` by docker by default
- fixed / unified some env handling

### DIFF
--- a/src/usr/local/bin/install-buildpack
+++ b/src/usr/local/bin/install-buildpack
@@ -28,7 +28,13 @@ fi
 
 # env helper, loads tool specific env
 cat >> $BASH_ENV <<- EOM
-export BUILDPACK=1 USER_NAME="${USER_NAME}" USER_ID="${USER_ID}"
+export BUILDPACK=1 USER_NAME="${USER_NAME}" USER_ID="${USER_ID}" USER_HOME="/home/${USER_NAME}"
+
+# openshift override unknown user home
+if [ "\${EUID}" != 0 ] && [ "\${HOME}" == "/" ]; then
+  export HOME="\${USER_HOME}"
+fi
+
 if [ -d /usr/local/env.d ]; then
   for i in /usr/local/env.d/*.sh; do
     if [ -r \$i ]; then
@@ -37,8 +43,8 @@ if [ -d /usr/local/env.d ]; then
   done
   unset i
 fi
-if [ -d /home/${USER_NAME}/env.d ]; then
-  for i in /home/${USER_NAME}/env.d/*.sh; do
+if [ -d ${USER_HOME}/env.d ]; then
+  for i in ${USER_HOME}/env.d/*.sh; do
     if [ -r \$i ]; then
       . \$i
     fi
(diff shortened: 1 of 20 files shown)


src/usr/local/bin/install-buildpack
```diff
@@ -28,7 +28,13 @@ fi
 
 # env helper, loads tool specific env
 cat >> $BASH_ENV <<- EOM
-export BUILDPACK=1 USER_NAME="${USER_NAME}" USER_ID="${USER_ID}"
+export BUILDPACK=1 USER_NAME="${USER_NAME}" USER_ID="${USER_ID}" USER_HOME="/home/${USER_NAME}"
+
+# openshift override unknown user home
+if [ "\${EUID}" != 0 ] && [ "\${HOME}" == "/" ]; then
```
rarkins: Home is `/` if run as a random ID?

src/usr/local/buildpack/tools/dotnet.sh
```diff
@@ -10,7 +10,7 @@ if [[ ! "${MAJOR}" || ! "${MINOR}" || ! "${PATCH}" ]]; then
   exit 1
 fi
 
-DOTNET_INSTALL_DIR=/usr/local/dotnet/${TOOL_VERSION}
+DOTNET_INSTALL_DIR=/usr/local/${TOOL_NAME}/${TOOL_VERSION}
```
rarkins, line 13: Is there a scenario where this is not `dotnet` ?

src/usr/local/buildpack/tools/elixir.sh
```diff
@@ -5,12 +5,17 @@ set -e
 require_root
 check_command erl
 
+if [[ -d "/usr/local/${TOOL_NAME}/${TOOL_VERSION}" ]]; then
+  echo "Skipping, already installed"
+  exit 0
+fi
```
rarkins, line 11: Is this specific to elixir? Seems like a generic rule. Easier copy/paste, less chance of human error?

src/usr/local/bin/install-buildpack
```diff
@@ -28,7 +28,13 @@ fi
 
 # env helper, loads tool specific env
 cat >> $BASH_ENV <<- EOM
-export BUILDPACK=1 USER_NAME="${USER_NAME}" USER_ID="${USER_ID}"
+export BUILDPACK=1 USER_NAME="${USER_NAME}" USER_ID="${USER_ID}" USER_HOME="/home/${USER_NAME}"
+
+# openshift override unknown user home
+if [ "\${EUID}" != 0 ]; then
+  export HOME="\${USER_HOME}"
+fi
```
viceice, line 36: this should makes it openshift compatible. So override home if we are not root

src/usr/local/bin/install-buildpack
viceice: yes, at least on docker. but removed that condition, so only root is checked

src/usr/local/buildpack/tools/dotnet.sh
viceice: not yet, just some simplification, so i can easily copy'n'past between tools 🙃

src/usr/local/buildpack/tools/elixir.sh
viceice: not yet really happy with this, as for some tools i like to support multiple versions and latest install call should override env again, so executables in path always points to latest installed version.

see python or php for a more generic version. will extract some more helper functions later to make new tools even more easy to implement.

rarkins (approved):

PR #94, opened 2021-07-15
feat: support ignore tools by arg

### LINKED ISSUE #54: Support `IGNORED_TOOLS` arg
We should support a `IGNORED_TOOLS` build argument, which will be a comma-separated list. If the list is non-empty, any tool within should be skipped for `install-tool`.

Example: `IGNORED_TOOLS=PIPENV,HELM`

Maybe we can make it case-insensitive? e.g. allowing `IGNORED_TOOLS=pipenv,helm` too.

viceice:
Allows to disable tool installation via build arg: `ARG IGNORED_TOOLS=powershell,node`.

closes #54

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,3 +55,25 @@ To use a custom base image with `containerbase/buildpack` checkout [custom-base-
 ### Custom Root CA Certificates
 
 To add custom root certifactes to the `containerbase/buildpack` base image checkout [custom-root-ca](./docs/custom-root-ca.md) docs.
+
+### Temporary disable tool installer
+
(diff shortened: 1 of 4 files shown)


rarkins (commented):
How does this work?? 
:)

rarkins (changes requested):
Needs documentation too

viceice:
`install-tool` simply checks `IGNORED_TOOLS` env variable. If the tool is listed there, it skips install.

rarkins (approved):

Issue #16, opened 2021-04-07
Allow runtime install of tools


viceice:
Make all possible tools runtime installable.

**TODO**
- [x] #121
- [x] #197
- [x] #213
- [x] `bundler` #251
- [x] `cocoapods`  #251
- [x] `composer`
- [x] `docker` #651
- [x] #118
- [x] `elixir` #376
- [x] #318
- [x] #200 
- [x] `git-lfs` #652
- [ ] ~`git`~ not yet possible
- [x] #119
- [x] `gradle`
- [x] `hashin` (`pip_requirements`) #639
- [x] #199
- [x] `java`
- [x] #178
- [x] `nix` #573
- [x] #217
- [x] `npm`
- [x] `pipenv` #639
- [x] `pip-tools` (`pip-compile`) #639
- [x] #120
- [x] `pnpm`
- [x] `poetry`
- [x] #633
- [x] #404
- [x] #475
- [x] #571
- [x] `sbt`
- [x] `scala`
- [x] #576
- [x] #181, #198
- [x] `yarn`

rarkins:
I don't think it really matters much  whether it's in `/home/user` or `/usr/local` however `/usr/local` would probably be better for OpenShift, where the UID is random.

viceice:
> We like to implement a new `binarySource=install` where renovate install tools dynamically at runtime.
> 
> Use-case is to have smaller docker image used in kubernets, which can download missing tools on demand.
> 
> ref #96

https://github.com/renovatebot/docker-buildpack/issues/144#issue-802657002

rarkins:
@viceice should we break these into separate issues to research and complete one by one? I'd like to know early if we have any "blockers", too

rarkins:
Thinking it through, we'll need:
1. OS pre-requisites which need to be built into the base image, mostly stuff that requires `apt`
1. "prebuilt" versions of most languages (we do not want to compile them at install time) and some tools. For these we do the prebuilding ourselves, currently publishing to GitHub Releases
1. most package manager tools dynamically installable at run-time, from their origin internet sources

So in terms of research, it's a matter of working out which falls in (2) and which in (3).

Pre-built:
- [ ] node?
- [x] php
- [x] python
- [x] ruby

Dynamically-installable:
- ?

(let's update the above lists as we go)

viceice:
No easy way beside to check every tool.
I think most tools can be installed as runtime if we use a file download.

[...]

PR #99, opened 2021-07-22
feat: add libmysqlclient-dev to python buildpack

wastrachan:


wastrachan:
Issues reference: https://github.com/containerbase/buildpack/issues/98

src/usr/local/buildpack/tools/python.sh
```diff
@@ -32,6 +32,11 @@ if [[ -z "${tool_path}" ]]; then
 
   curl -sSfLo ${file} ${PYTHON_URL}/${TOOL_VERSION}/python-${TOOL_VERSION}-${CODENAME}-${ARCH}.tar.xz || echo 'Ignore download error'
 
+  apt_install \
+    build-essential \
+    libmysqlclient-dev \
+    ;
```
viceice, line 38: This will not work as is, because some tests are already running without root, so they will now fail. We need some abstractions for installing base deps before we can add this.

See the other open issues for more infos

viceice (changes requested):

viceice:
Depends on #16

viceice:
Ref #30

PR #135, opened 2021-09-11
docs: add USER_NAME etc

rarkins:


docs/custom-base-image.md
```diff
@@ -12,10 +12,17 @@ You should always use a specific version which can be found at [docker hub](http
...
+# This default may be swapped for any compatible base image
+ARG BASE_IMAGE=amd64/ubuntu:20.04
+
 # This buildpack is used for tool intallation and user/directory setup
 FROM containerbase/buildpack AS buildpack
 
-FROM amd64/ubuntu:focal as base
+FROM ${BASE_IMAGE} as base
+
+ARG USER_NAME=ubuntu
```
viceice, line 23: Did you read the description above the sample? It was intended to not override User here.

rarkins: I had the build fail without it. Will try again

viceice: If it won't work there's a bug

PR #218, opened 2021-12-07
feat(node): allow runtime install

### LINKED ISSUE #217: node: runtime install
`node` (tricky, to not interferre with renovate required node)

viceice:
- closes #217

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -16,6 +16,11 @@ variable "CACHE_WEEK" {
   default = ""
 }
 
+
+variable "BUILDPACK_DEBUG" {
+  default = ""
+}
+
 group "default" {
   targets = ["build-docker"]
 }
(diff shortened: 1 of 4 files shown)


rarkins:
Do we need to purge any node installs (e.g. npm, pnpm, yarn) if we change node version?

viceice:
> Do we need to purge any node installs (e.g. npm, pnpm, yarn) if we change node version?

I don't think so. Why should we need to purge?

rarkins:
Might any of them build/install differently based on the version of node? I guess I'm thinking about re2 and similar

viceice:
I've never seen that behavior on package managers, but i can add some more tests to verify

rarkins:
We'd need it later for Eg new python means purging any pop or poetry?

viceice:
Maybe, let's defer it, until we have something to test against.

PR #238, opened 2022-01-05
fix: properly test for installed git-lfs

gjasny:
The test before claimed that `git-lfs` is already installed, even if it was not.

### DIFF
--- a/src/usr/local/buildpack/tools/git-lfs.sh
+++ b/src/usr/local/buildpack/tools/git-lfs.sh
@@ -16,9 +16,9 @@ if [[ -x "$(command -v git-lfs)" ]]; then
 fi
 
 ARCH=linux-amd64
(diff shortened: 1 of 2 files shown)


viceice:
Can you describe, why the old test is not working as expected?

gjasny:
> Can you describe, why the old test is not working as expected?

Because no test existed. I now invoke `git lfs version` after installation. Before my change it failed as could be seen [here](https://github.com/gjasny/buildpack/runs/4729162660?check_suite_focus=true).

viceice (commented):
Ci still fails

gjasny:
> Ci still fails

CI on my branch is green: https://github.com/gjasny/buildpack/actions/runs/1667814918

viceice (approved):

PR #252, opened 2022-01-21
feat(ruby): dynamic bundler and cocoapods

### LINKED ISSUE #251: Support install-tool for Bundler and Cocoapods
Currently these are both installed using `install-gem`

viceice:
closes #251

### DIFF
--- a/src/usr/local/buildpack/tools/bundler.sh
+++ b/src/usr/local/buildpack/tools/bundler.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+check_command ruby
+
+. /usr/local/buildpack/utils/ruby.sh
+
+gem_install
+gem_shell_wrapper
+
+bundler --version
--- a/test/ruby/Dockerfile
+++ b/test/ruby/Dockerfile
@@ -30,9 +30,9 @@ COPY --chown=1000:0 test test
 WORKDIR /test
 
 #--------------------------------------
-# test: bundler
+# test: bundler (gem)
 #--------------------------------------
-FROM build as testa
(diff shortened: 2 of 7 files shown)


src/usr/local/buildpack/tools/bundler.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+check_command ruby
+
+. /usr/local/buildpack/utils/ruby.sh
+
+gem_install
+gem_shell_wrapper
```
rarkins, line 10:
```suggestion
gem_shell_wrapper bundler
```

?

I see a difference between this and `pod` below

test/ruby/Dockerfile
```diff
@@ -52,12 +52,12 @@ RUN set -ex; \
 
 RUN set -ex; \
     cd a; \
-    bundle lock
+    bundler lock
```
rarkins, line 55: Can you refresh me on `bundle` vs `bundler`?

viceice: It's the same, renovate uses `bundler` like the package name, `bundle` is the older name.

viceice:
`/usr/local/cocoapods/1.11.2/gems/claide-1.1.0/lib/claide/command.rb:439:in `help!': \e[31m[!] You cannot run CocoaPods as root.\e[39m (CLAide::Help)` 

🤦‍♂️

src/usr/local/buildpack/tools/bundler.sh
viceice: Yes, the arg is only required when different to tool name. 😉

PR #253, opened 2022-01-22
feat: runtime install of golang

### LINKED ISSUE #119: golang: runtime install
- [x] extract prepare-tool
- [x] allow user install

Chumper:
Enables runtime installation of golang.
Installation seems fine for non root install.

Local tests also work.

Closes #119

### DIFF
--- a/src/usr/local/buildpack/tools/golang.sh
+++ b/src/usr/local/buildpack/tools/golang.sh
@@ -2,7 +2,6 @@
 
 set -e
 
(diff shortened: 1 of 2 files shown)


viceice (changes requested):
It's installing bazar, that needs root. So we need the prepare feat first.

Chumper:
Will include this changes in the `prepare-tool` feature where I will use golang as example

PR #255, opened 2022-01-23
feat: prepare tool for golang

### LINKED ISSUE #119: golang: runtime install

### LINKED ISSUE #121: prepare-tool: install tool base deps as root
Add a `prepare-tool` to install all required distro base dependencies (apt).

Chumper:
Depends on #254 which needs to be merged first, before we can take a look at this PR.

Adds a `prepare-tool` script that will be used to make tools run time install-able.
A draft first to gather feedback and make sure we are moving into the right direction.

- Closes #119 
- Closes #121

### DIFF
--- a/src/usr/local/bin/prepare-tool
+++ b/src/usr/local/bin/prepare-tool
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -e
+
+. /usr/local/buildpack/util.sh
+
+require_distro
+require_user
+require_tool "$@"
+
+PREPARE_TOOL="/usr/local/buildpack/prepare/${TOOL_NAME}.sh"
+if [[ ! -f "$PREPARE_TOOL" ]]; then
+  echo "No tool defined - skipping: ${TOOL_NAME}" >&2
+  exit 1;
+fi
+
+if [[ $(ignore_tool) -eq 1 ]]; then
+  echo "Tool ignored - skipping: ${TOOL_NAME}"
+  exit 0;
+fi
+
+
+echo "Preparing tool ${TOOL_NAME}"
+
+. "$PREPARE_TOOL"
+
+# cleanup
+if [[ $EUID -eq 0 ]]; then
+  rm -rf /var/lib/apt/lists/* /tmp/*
+fi
--- a/src/usr/local/bin/prepare-tool
+++ b/src/usr/local/bin/prepare-tool
@@ -5,7 +5,7 @@ set -e
 . /usr/local/buildpack/util.sh
 
 require_distro
-require_user
+require_root
 require_tool "$@"
 
 PREPARE_TOOL="/usr/local/buildpack/prepare/${TOOL_NAME}.sh"
--- a/src/usr/local/bin/prepare-tool
+++ b/src/usr/local/bin/prepare-tool
@@ -2,6 +2,7 @@
 
 set -e
 
+# shellcheck source=/dev/null
 . /usr/local/buildpack/util.sh
 
 require_distro
@@ -11,7 +12,7 @@ require_tool "$@"
 PREPARE_TOOL="/usr/local/buildpack/prepare/${TOOL_NAME}.sh"
 if [[ ! -f "$PREPARE_TOOL" ]]; then
   echo "No tool defined - skipping: ${TOOL_NAME}" >&2
-  exit 1;
+  exit 0;
 fi
 
 if [[ $(ignore_tool) -eq 1 ]]; then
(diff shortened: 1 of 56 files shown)


src/usr/local/buildpack/utils/ruby.sh
```diff
@@ -4,36 +4,38 @@ function gem_install() {
...
+    GEM_HOME=$tool_path gem install  --bindir "$tool_path"/bin "${TOOL_NAME}" -v "${TOOL_VERSION}"
 
   # TODO: clear gem cache
   fi
 }
 
 
 function gem_link_wrapper() {
-  link_wrapper ${1:-$TOOL_NAME} $tool_path/bin
+  link_wrapper "${1:-$TOOL_NAME}" "$tool_path/bin"
```
viceice, line 18: Looks promising, but can we split out those changes as small fix Pr's, so we have less changes here?

test/golang/Dockerfile
```diff
@@ -43,11 +46,9 @@ FROM build as testb
 RUN install-tool golang 1.11.0
 
 RUN set -ex; \
-    cd d; \
-    go get -d ./...; \
-    go mod tidy;
-
-USER 1000
+  cd d; \
```
viceice, line 49: What's causing those indention changes?

Chumper: My prettifyer... I'll change it back

viceice: Do i missed to add a prettier config? 🤔

We should have it to suppress those changes in future. 🤗

src/usr/local/bin/prepare-tool
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+set -e
+
+. /usr/local/buildpack/util.sh
+
+require_distro
+require_root
+require_tool "$@"
```
viceice, line 9: This probably fails when no version is found

https://github.com/containerbase/buildpack/blob/48c5719e1537bc1c2abc4466a31fb46a831dcade/src/usr/local/buildpack/util.sh#L214-L217

So we maybe need to refactor / split that function too.

viceice:
Ups, i accidently reverted the `USER 1000` remove line

src/usr/local/bin/prepare-tool
viceice: Strange, it didn't failed as expected 😕

src/usr/local/buildpack/prepare/golang.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -e
+
+require_root
+
```
viceice, line 6: Should already be be enforced
```suggestion
```

viceice (commented):
Some thoughts:
* we should test install as root
* if running `install-tool` as root, `prepare-tool` should be run implicit for easier/ smaller dockerfiles
* `prepare-tool` should skip duplicate installs, i think we can create the tool dir and check if it's exist.

src/usr/local/bin/prepare-tool
Chumper: It did not, but I will investigate, because I had the same thought

src/usr/local/buildpack/prepare/golang.sh
viceice: yes, i think it's easier to check if the `golang` folder at `/usr/local/golang` exists than checking if all deb packages are created.

On some tools we make some more configs on prepare, so i think the simplest check is to validate the folder exists.


I recently had the idea to simplify folder handling.
* install all tools to `/usr/local` (even user install)
* allow root group writing to eg `/usr/local/golang` and `/usr/local/bin`
* should make prepare and install checks more easy.
WDYT?

Chumper: > yes, i think it's easier to check if the `golang` folder at `/usr/local/golang` exists than checking if all deb packages are created.
>
> On some tools we make some more configs on prepare, so i think the simplest check is to validate the folder exists.

Fair point

> I recently had the idea to simplify folder handling.
> 
>     * install all tools to `/usr/local` (even user install)
> 
>     * allow root group writing to eg `/usr/local/golang` and `/usr/local/bin`
> 
>     * should make prepare and install checks more easy.
>       WDYT?

This would indeed make it easier.
We can handle root installs with the root group and make all tools owned by the 1000 user.

Chumper: We should make the unified install directory before this PR then.

viceice: tool versions installed via root should not be changeble by user. so only the directory should be writable for new  files

Chumper:
Will use a different approach, making the tool files sourcable by providing interface kind of functions

PR #258, opened 2022-01-24
fix: shellcheck issues

viceice:
- #256

PR #263, opened 2022-01-25
fix: shellcheck errors

Chumper:
Builds on top of #258 

Solves the remaining shellcheck errors

### DIFF
--- a/src/usr/local/buildpack/tools/python.sh
+++ b/src/usr/local/buildpack/tools/python.sh
@@ -22,6 +22,7 @@ if [[ -z "${tool_path}" ]]; then
   file=/tmp/python.tar.xz
 
   ARCH=$(uname -p)
+  # shellcheck source=/dev/null
   CODENAME=$(. /etc/os-release && echo "${VERSION_CODENAME}")
   PYTHON_URL="https://github.com/containerbase/python-prebuild/releases/download"
 
@@ -56,7 +57,9 @@ if [[ -z "${tool_path}" ]]; then
   fi
 
   fix_python_shebangs() {
-    for file in $(find "${tool_path}/bin" -type f -exec grep -Iq . {} \; -print); do
+    # https://github.com/koalaman/shellcheck/wiki/SC2044
+    while IFS= read -r -d '' file
+    do
       case "$(head -1 "${file}")" in
       "#!"*"/bin/python" )
         sed -i "1 s:.*:#\!${tool_path}\/bin\/python:" "${file}"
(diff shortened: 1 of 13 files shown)


src/usr/local/buildpack/tools/python.sh
```diff
@@ -22,6 +22,7 @@ if [[ -z "${tool_path}" ]]; then
   file=/tmp/python.tar.xz
 
   ARCH=$(uname -p)
+  # shellcheck source=/dev/null
   CODENAME=$(. /etc/os-release && echo "${VERSION_CODENAME}")
```
viceice, line 26: We should extract this to a util function

src/usr/local/buildpack/tools/swift.sh
```diff
@@ -14,13 +14,15 @@ base_path=/usr/local/buildpack/${TOOL_NAME}
 tool_path=$base_path/$TOOL_VERSION
 
 if [[ ! -d "${tool_path}" ]]; then
+  # shellcheck source=/dev/null
   VERSION_CODENAME=$(. /etc/os-release && echo "${VERSION_CODENAME}")
-  VERSION_ID=$(. /etc/os-release && echo "${VERSION_ID}")
+  # shellcheck source=/dev/null
+  VERSION_CODENAME=$(. /etc/os-release && echo "${VERSION_ID}")
```
viceice, line 20:
```suggestion
  VERSION_ID=$(. /etc/os-release && echo "${VERSION_ID}")
```

viceice:
Looks like we need some lowercase stuff on workflow. Can we do this with an expression? 🤔

src/usr/local/buildpack/tools/python.sh
Chumper: I can do that, we have the utils file included everywhere.

Chumper:
> Looks like we need some lowercase stuff on workflow. Can we do this with an expression? 🤔

What do you mean exactly?

src/usr/local/buildpack/util.sh
```diff
@@ -190,6 +190,12 @@ function require_distro () {
   esac
 }
 
+function get_distro() {
+  require_distro
+  # shellcheck source=/dev/null
+  . /etc/os-release && echo "${VERSION_CODENAME}"
```
viceice, line 196: I think we need a subshell here, otherwise we get all variables in our scope.
```suggestion
  echo $(. /etc/os-release && echo "${VERSION_CODENAME}")
```

viceice (commented):
See the failing ci, it's complaining about wrong docker image casing. We use the organization name and yours contains uppercase.

viceice (approved):

PR #269, opened 2022-02-01
feat: adding defaults for tool handling

Chumper:
This contains major refactorings regarding:
* tool handling
* paths
* tests

This is a draft to get general feedback during implementation and possibly to split of tasks that can be done independently

# Concept

The following sections describe the concepts of the refactoring

## Tool Handling

The most integral part of the refactoring is the way how tools are installed and managed.

Previously the tools were simple bash files that were sourced and installed the tool during the source.  
The tool was responsible for detecting if it has been installed and the complete setup.

This will change to a more function based approach.

Each tool needs to provide the following function in the tool file that will be sourced:

* `check_tool_requirements`
  > This function will be used to check if the tool can be installed with the current user and the current version.
* `check_tool_installed`
  > This function will check if the tool in the given version is already installed.
* `install_tool`
  > This function will install the tool in the given version.
* `link_tool`
  > This function will link the tool in the given version into a usable binary.
* `prepare_tool`
  > This function will be called as root and should install needed packages that are required to make the tool installable during runtime.

With this set of functions we can guarantee that the tool handling can be mainstreamed for general use.

It also makes it easier to implement caching of downloaded tools.

## Paths

Previously tools have been installed into `/usr/local/...` and the users home folder (typically `/home/user/bin`).  
We are now differentiating between scripts provided by the `buildpack` image which are still located at `/usr/local/...` and the tools, environment and versions that are getting installed.

We are also using a single file path for all installed tools and make use of owner and group membership to prevent `user`s installable tools to overwrite `root` installed tools.

The current path for installed tools is `/opt/buildpack/...`  
The following folders exist:

* `/opt/buildpack/tools/`
  > Contains all install and runtime files 
* `/opt/buildpack/env.d/`
  > Contains all environment specific settings for the tools
* `/opt/buildpack/versions/`
  > Contains a file for each tool with the current installed version
* `/opt/buildpack/bin/`
  > Contains linked wrappers for each tool

## Tests

I have moved as many operations as possible into the `util.sh` file and enhanced and refactored it.
This enables us to make changes in a central place and affect all tools, e.g. it is now simple to change the global install directory.

I have also moved the test files to make testing using bats possible for the bash files.
This offers us to test the tools in isolation as well as with the docker integration tests.


src/usr/local/buildpack/utils/defaults.sh
```diff
@@ -0,0 +1,36 @@
...
+}
+
+# Links the tools installation to the global bin folders
+function link_tool () {
+  echo "'link_tool' not defined for tool ${TOOL_NAME}"
+  exit 1
+}
+
+# Installs needed packages to make the tool runtime installable
+function prepare_tool() {
```
viceice, line 33: I think this can silently exit, as not all tools need a prepare phase?

src/usr/local/buildpack/utils/folders.sh
```diff
@@ -0,0 +1,145 @@
...
+  if [ -d "${folder}" ]; then
+    return
+  fi
+
+  if [ ! -d "${parent}" ]; then
+    create_folder "$parent"
+  fi
+
+  local umask=$USER_UMASK
+  if [ "$(is_root)" -eq 0 ]; then
```
viceice, line 43: `is_root`? Do we need this function?

src/usr/local/buildpack/utils/folders.sh
```diff
@@ -0,0 +1,145 @@
...
+    umask=$ROOT_UMASK
+  fi
+
+  mkdir -m "${umask}" "${folder}"
```
viceice, line 47: 🤔 this way the user can't install a new node version when one was already installed as root?

I think we need to be able to do that.

viceice (commented):
Looks promising, will definitely be a big breaking change.

Thinking about how to get some compatibility for existing users who simply put additional tools in extended docker images. Eg our helm charts repo.

Can we use the old tools folder for compabillity and use a new folder for the new stuff? 🤔

src/usr/local/buildpack/utils/folders.sh
Chumper: It is needed to make testing possible. 
I also changed the `utils.sh` but I am struggling to make commits that will not brake anything.
So locally it is already changed and it makes sense as soon as I push more code :)

src/usr/local/buildpack/utils/folders.sh
Chumper: Correct, we only want to protect root installed versions, not that the user cannot install anything. will change it.

Chumper:
@viceice I think is the minimal change that we can introduce.

Summary of what I have done:

* Moved all tools into a `legacy_tool_install` function
* The `install-tool` will check if the tool is legacy and uses the legacy function to install it.
  This makes it backwards compatible and we can move tool for tool to the new way of installing tools.
* Split up the utils function into multiple smaller pieces and made them testable.
* Prepared all kind of code to replace the legacy code in the `/usr/local/bin/` directory.

36 files are the tool files that were moved into functions.

After cleaning up this PR, clarifying understandings and agreeing on certain standards we can merge this and make smaller incremental changes to the tools and the whole install process.

As far as I can see this PR is NOT breaking, I haven't touched the docker tests yet.

src/usr/local/buildpack/utils/folders.sh
Chumper: https://github.com/containerbase/buildpack/pull/269/files#diff-a968fbd6e0e6b76f847072007c8b5d39bfcbd8f06e8d8663007b394bf131a3ecR45-R48

Here you see it is needed to test the files with "root" id and without

src/usr/local/buildpack/utils/folders.sh
Chumper: Haven't changed this yet, we need to keep this in mind when we are migrating the tools and writing tests for them

Chumper:
Actions on my fork are all green: https://github.com/Chumper/buildpack/actions/runs/1781263315

viceice:
Would be nice to not break this

https://github.com/renovatebot/helm-charts/tree/main/.devcontainer

Chumper:
Understand, let me split this PR up in smaller parts to get the initial ball rolling:

* tools in `/usr/local/buildpack/tools` are considered legacy and are installed the current way.
  So the first PR will introduce the tools "contract" with the `defaults.sh` for tools in a new location. 
  No changes to tools at all.
* Splitting up the util.sh into smaller files, no changes to code, just moving functions
* Adding tests via bats for the helper files
* Refactoring of util functions (I found some variable check that should be changed in my opinion, open to discussion)
* Refactoring one example tool to the new way

Subsequentially we can implement caching, sub language installation in the "new" tools while still being backwards compatible

viceice:
Sounds good

Chumper:
Closing this in favor of multiple smaller PRs.

PR #271, opened 2022-02-02
feat: add new tool contract

Chumper:
This adds the new tools contract only.

When a tool is located in `/usr/local/buildpack/tools/v2/<TOOL>.sh` then it is considered a new tool (in terms of contract) and will be installed with the methods defined in the `defaults.sh`.

When the tool is available in the old and new path, then the new path is preferred.


viceice (approved):

viceice:
Shouldn't this be a new feature than a chore?

Chumper:
Yes, should be indeed a feature instead of a chore. 
Just because we have no new tools it still could be used as new feature.

PR #274, opened 2022-02-03
test: add bats test suite to utility files

Chumper:
Adds test for the current util files:

- [x] `util.sh`
- [x] `environment.sh`
- [x] `filesystem.sh`
- [x] `linking.sh`

Not sure why it replaces the versions in the `yarn.lock`. If you know how to fix this, let me know.
I added bats with `yarn add --dev <url>`


test/bash/util.bats
```diff
@@ -0,0 +1,158 @@
...
+@test "require_tool" {
+  run require_tool
+  assert_output --partial "No tool defined"
+  assert_failure
+
+  # TODO(Chumper): In my opinion this test should fail
+  run require_tool foo
+  # assert_output --partial "No version defined"
+  # assert_failure
+  assert_success
```
Chumper, line 26: @viceice I would like your opinion on this. 
Is this working as intended?

test/bash/util.bats
```diff
@@ -0,0 +1,158 @@
...
+@test "requires_user" {
+  run require_user
+  assert_output --partial "No USER_NAME defined"
+  assert_failure
+
+  # TODO(Chumper): In my opinion this test should fail
+  USER_NAME= run require_user
+  # assert_output --partial "No USER_NAME defined"
+  # assert_failure
+  assert_success
```
Chumper, line 51: @viceice Same here, I would like your opinion on this. 
Is this working as intended?

test/bash/util.bats
```diff
@@ -0,0 +1,158 @@
...
+
+@test "check_command" {
+  run check_command foo
+  assert_failure
+  assert_output --partial "No foo defined"
+
+  # TODO(Chumper): This command should fail in my opinion
+  run check_command
+  assert_failure
+  assert_output --partial "No  defined"
```
Chumper, line 114: @viceice Same here, I would like your opinion on this. 
Is this working as intended?

test/bash/util.bats
```diff
@@ -0,0 +1,158 @@
...
+  # Not testable that easily, needs to be refactored first
+}
+
+@test "ignore_tool" {
+
+  # TODO(Chumper): Should this succed?
+  TOOL_NAME= \
+  IGNORED_TOOLS= \
+  run ignore_tool
+  assert_output "1"
```
Chumper, line 142: @viceice Same here, I would like your opinion on this. 
Is this working as intended?

Chumper:
@viceice We broke the `shell_wrapper` function with our shellcheck changes: https://github.com/containerbase/buildpack/pull/258/files#diff-01d7a5c53bd60ed4d24d3a6a36b3450bdcab4927ba6a9b487e83811ee87f4ae1R90

I should be able to fix it once I am refactoring the utils files.

Chumper:
Should be ready for review. I left a bit of code in it that I will enable once we have testable methods.

test/bash/util.bats
viceice: I think it's ok, as we validate `TOOL_NAME` before?

test/bash/util.bats
viceice: i think this should fail

test/bash/util.bats
viceice: should fail

test/bash/util.bats
```diff
@@ -0,0 +1,159 @@
...
+
+@test "check_semver" {
+
+  run check_semver
+  assert_failure
+
+  run check_semver foo.bar.baz
+  assert_failure
+
+  check_semver 1.2.3
```
viceice, line 82: Test `11.0.14+9` too, just to be safe

https://github.com/containerbase/buildpack/blob/b09200b22e93dd61140d155a6e3c9c215b6ad5cd/test/java/Dockerfile#L19

test/bash/util.bats
viceice: yes, should fail

test/bash/util.bats
```diff
@@ -0,0 +1,159 @@
...
+  assert_output --partial "No  defined"
+
+  run check_command ls
+  assert_success
+}
+
+@test "require_distro" {
+  if ! check_command "lsb_release" || [ ! $(lsb_release -si) =~ "Ubuntu" ]; then
+    skip "not ubuntu"
+  fi
```
viceice, line 124: Are you testing on other systems?

Chumper: I have Windows WSL2 with ubuntu, but the devcontainer I used was Debian.
I am not sure if it safe to assume that all developers will be on ubuntu. 
So if you execute it locally it should skip the ubuntu only tests.

viceice: OK, so we should preprare a ubuntu devcontainer for local testing? I'm using WSL ubuntu too 😅

package.json
```diff
@@ -17,8 +17,11 @@
     "@semantic-release/exec": "6.0.3",
     "@types/node": "16.11.21",
     "@types/shelljs": "0.8.11",
-    "conventional-changelog-conventionalcommits": "4.6.3",
+    "bats": "^1.5.0",
+    "bats-assert": "https://github.com/ztombol/bats-assert",
+    "bats-support": "https://github.com/ztombol/bats-support",
```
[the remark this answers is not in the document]
Chumper: We should, still a npm newb, will change it

test/bash/util.bats
Chumper: That is still a PR I have open :)

test/bash/util.bats
Chumper: yes we do check the tool before

viceice (approved):

Chumper:
@viceice Is there a way to rerun the checks or do I need to push a bogus commit?

viceice:
> @viceice Is there a way to rerun the checks or do I need to push a bogus commit?

Retrigger seems to be hung, yesterday, so now done again. 🤔

PR #281, opened 2022-02-05
refactor: add feature flag implementation

Chumper:
Requires #274 to be merged first.
Will be rebased on the master then.

Adds a simple feature flag implementation that will be used to make old and new tools coexist.
Changes are in https://github.com/containerbase/buildpack/commit/580f54ec8dd5af9954808d40df05623d897aa15f


viceice:
Needs deconfliction 🤗

Chumper:
I want to put this back for a while. 

I originally planed to us the feature flags to determine if we need the legacy behavior or the new v2 behavior.
I thought about it yesterday and I think we can reach the same functionality by having method overwrites.

e.g. the tests will overwrite the `get_install_dir` functions and all others where we need to inject a temporary test path.
We can do the same with v2 function, where we simply source a `src/usrl/local/buildpack/features/v2_tool.sh` which overwrites utils function needed for the new way of installing tools.
This makes testing pretty easy as everything is encapsulated in their own function.
Is there something you would prefer?

PR #283, opened 2022-02-06
test: refactor filesystem tests

Chumper:
This adds the the `is_root` methods to make the helper functions testable

Refactored the filesystem tests to be testable for both users.


src/usr/local/buildpack/utils/environment.sh
```diff
@@ -34,7 +34,7 @@ function reset_tool_env () {
 function find_tool_env () {
   local install_dir
   install_dir=$(get_install_dir)
-  if [[ -z "${TOOL_NAME+x}" ]]; then
+  if [[ -z "${TOOL_NAME}" ]]; then
```
viceice, line 37: So this should be a fix release?

viceice (commented):
Otherwise LGTM

src/usr/local/buildpack/utils/environment.sh
Chumper: We can do that if you want, it is just something I forgot when I added the tests in the last commit.
Effectively it changes this test: https://github.com/containerbase/buildpack/pull/283/files#diff-b43f91339e00e49afed5160a8bf590aff58dd2c160d06d363a840b2f5a378b0dL84-L87

PR #284, opened 2022-02-06
fix: proper validation of user variables

Chumper:
Extracts the `USER_HOME` variable to make it changeable during tests
Uses new test setup for environment and linking tests
Fix function that works even though it should be broken.

### DIFF
--- a/src/usr/local/buildpack/utils/environment.sh
+++ b/src/usr/local/buildpack/utils/environment.sh
@@ -79,7 +79,7 @@ function setup_env_files () {
 export BUILDPACK=1 USER_NAME="${USER_NAME}" USER_ID="${USER_ID}" USER_HOME="${USER_HOME}"
 
 # openshift override unknown user home
-if [ "$(is_root)" -ne 0 ]; then
+if [[ $EUID -ne 0 ]]; then
   export HOME="\${USER_HOME}"
 fi
 
--- a/src/usr/local/buildpack/utils/environment.sh
+++ b/src/usr/local/buildpack/utils/environment.sh
@@ -34,7 +34,7 @@ function reset_tool_env () {
 function find_tool_env () {
   local install_dir
   install_dir=$(get_install_dir)
-  if [[ -z "${TOOL_NAME+x}" ]]; then
+  if [[ -z "${TOOL_NAME}" ]]; then
     echo "No TOOL_NAME defined - skipping: ${TOOL_NAME}" >&2
     exit 1;
   fi
--- a/src/usr/local/buildpack/utils/environment.sh
+++ b/src/usr/local/buildpack/utils/environment.sh
@@ -77,7 +77,7 @@ function setup_env_files () {
 export BUILDPACK=1 USER_NAME="${USER_NAME}" USER_ID="${USER_ID}" USER_HOME="${USER_HOME}"
 
 # openshift override unknown user home
-if [[ $EUID -ne 0 ]]; then
+if [[ "\${EUID}" -ne 0 ]]; then
   export HOME="\${USER_HOME}"
 fi
 
--- a/src/usr/local/buildpack/utils/environment.sh
+++ b/src/usr/local/buildpack/utils/environment.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-export ENV_FILE=/usr/local/etc/env
-
 function refreshenv () {
   if [[ -r "$ENV_FILE" ]]; then
     # shellcheck source=/dev/null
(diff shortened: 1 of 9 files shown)


Chumper:
Need to see that I will not use functions in the bash_env :(

Chumper:
Reverted root function usage in the env file. it is only used for Openshift and not important for testing.

Chumper:
Took too long to find the quotation error...

src/usr/local/buildpack/utils/environment.sh
```diff
@@ -72,24 +70,27 @@ function get_tool_version_env () {
...
+  local install_dir
+  install_dir=$(get_install_dir)
+
   cat >> "$ENV_FILE" <<- EOM
-export BUILDPACK=1 USER_NAME="${USER_NAME}" USER_ID="${USER_ID}" USER_HOME="/home/${USER_NAME}"
+export BUILDPACK=1 USER_NAME="${USER_NAME}" USER_ID="${USER_ID}" USER_HOME="${USER_HOME}"
 
 # openshift override unknown user home
-if [ "\${EUID}" != 0 ]; then
+if [[ "\${EUID}" -ne 0 ]]; then
```
viceice, line 80: So this now only works in bash? THis will thow syntax errors, when running default ubuntu dash

src/usr/local/buildpack/utils/environment.sh
```diff
@@ -72,24 +70,27 @@ function get_tool_version_env () {
...
   export HOME="\${USER_HOME}"
 fi
 
-if [ -d /usr/local/env.d ]; then
-  for i in /usr/local/env.d/*.sh; do
+if [ -d "${install_dir}/env.d" ]; then
+  for i in "${install_dir}/env.d/"*.sh; do
```
viceice, line 85: I think this should be hard coded. this can't change after buildpack install.
```suggestion
if [ -d "${ROOT_DIR}/env.d" ]; then
  for i in "${ROOT_DIR}/env.d/"*.sh; do
```

Or is this replaced before writing to file?

Chumper: This is replaced when writing to the file, otherwise I would also need to quote it.
Reason for extracting is so that it can be tested.

Chumper:
```
export BUILDPACK=1 USER_NAME="user" USER_ID="1000" USER_HOME="/home/user"

# openshift override unknown user home
if [ "${EUID}" != 0 ]; then
  export HOME="${USER_HOME}"
fi

if [ -d "/usr/local/env.d" ]; then
  for i in "/usr/local/env.d/"*.sh; do
    if [ -r $i ]; then
      . $i
    fi
  done
  unset i
fi
if [ -d "/home/user"/env.d ]; then
  for i in "/home/user"/env.d/*.sh; do
    if [ -r $i ]; then
      . $i
    fi
  done
  unset i
fi
export PATH="/home/user/bin:${PATH}"
export DEBIAN_FRONTEND=${DEBIAN_FRONTEND-noninteractive}
export LC_ALL=${LC_ALL-C.UTF-8}
export LANG=${LANG-C.UTF-8}
```
This is what it looks like in the docker image

viceice: `"/home/user"/env.d` looks ugly, should be changed to `"/home/user/env.d"`

src/usr/local/buildpack/utils/environment.sh
```diff
@@ -72,24 +70,27 @@ function get_tool_version_env () {
...
     if [ -r \$i ]; then
       . \$i
     fi
   done
   unset i
 fi
-if [ -d /home/"${USER_NAME}"/env.d ]; then
-  for i in /home/"${USER_NAME}"/env.d/*.sh; do
+if [ -d "${USER_HOME}"/env.d ]; then
+  for i in "${USER_HOME}"/env.d/*.sh; do
```
viceice, line 93:
```suggestion
if [ -d "${USER_HOME}/env.d" ]; then
  for i in "${USER_HOME}/env.d/"*.sh; do
```

Would  `"${USER_HOME}/env.d/*.sh"` work too? 🤔

Chumper: It would not, because then the glob pattern would not work: https://github.com/koalaman/shellcheck/wiki/SC2086

```
When quoting composite arguments, make sure to exclude globs and brace expansions, which lose their special meaning in double quotes: "$HOME/$dir/src/*.c" will not expand, but "$HOME/$dir/src"/*.c will.
```

viceice: OK, updated suggestion. 🙃

viceice (approved):

PR #285, opened 2022-02-07
feat: path can be appended to the end

Chumper:
* Extracted the bashrc location to the constants
* Enabled more in deep tests that require the constants to be changeable
* small indentation format

### DIFF
--- a/src/usr/local/buildpack/utils/environment.sh
+++ b/src/usr/local/buildpack/utils/environment.sh
@@ -99,7 +99,7 @@ if [ -d "${USER_HOME}/env.d" ]; then
 fi
 EOM
 
-  cat >> /etc/bash.bashrc <<- EOM
+  cat >> "${BASH_RC}" <<- EOM
 if [[ -r "$ENV_FILE" && -z "${BUILDPACK+x}" ]]; then
   . $ENV_FILE
 fi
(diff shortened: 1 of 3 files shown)


src/usr/local/buildpack/utils/environment.sh
```diff
@@ -52,14 +52,24 @@ function export_tool_env () {
...
+  if [[ -z "${TOOL_NAME}" ]]; then
     echo "No TOOL_NAME defined - skipping: ${TOOL_NAME}" >&2
     exit 1;
   fi
-  export PATH="$1:$PATH"
-  echo export PATH="$1:\$PATH" >> "$install_dir"/env.d/"${TOOL_NAME}".sh
+
+  if [ "${add_to_end}" = true ]; then
+    export PATH="$PATH:$additional_path"
+    echo export PATH="\$PATH:$additional_path" >> "${install_dir}/env.d/${TOOL_NAME}".sh
```
viceice, line 68: I think we should move the quote to end here

src/usr/local/buildpack/utils/environment.sh
```diff
@@ -52,14 +52,24 @@ function export_tool_env () {
...
+  else
+    export PATH="$additional_path:$PATH"
+    echo export PATH="$additional_path:\$PATH" >> "${install_dir}/env.d/${TOOL_NAME}".sh
```
viceice, line 71: And here

viceice (approved):

PR #293, opened 2022-02-10
refactor: add version helper with tests

Chumper:
This adds support to set versions for the tools and to query them
This is useful to detect if we need to link the tools again and can possibly help with further Renovate use cases.


viceice (commented):
otherwise LGTM

viceice (approved):

PR #294, opened 2022-02-10
refactor: add a cache implementation for downloads

### LINKED ISSUE #197: Cache dir / volume
We should support an optional cache dir /volume to cache downloads, so we can mount a host volume / docker volume and reuse downloads to speedup later runs.

So Cache will be enabled if `BUILDPACK_CACHE` points to an existing folder.

Chumper:
Extracts the constants into its own file
Adds a cache implementation that can be used to cache downloaded files

Partially closes #197, at least for new tools that will use the functions

### DIFF
--- a/src/usr/local/buildpack/utils/cache.sh
+++ b/src/usr/local/buildpack/utils/cache.sh
@@ -8,16 +8,21 @@ function get_from_url () {
   local url=${1}
   check url true
 
+  local checksum
+  checksum=$(echo "${url}" | sha1sum | awk '{print $1}')
+
   local name
   name=${2:-$(basename "${url}")}
 
-  if [ -n "${BUILDPACK_CACHE_DIR}" ] && [ -e "${BUILDPACK_CACHE_DIR}/${name}" ]; then
+  local filename="${checksum}/${name}"
+
+  if [ -n "${BUILDPACK_CACHE_DIR}" ] && [ -e "${BUILDPACK_CACHE_DIR}/${filename}" ]; then
       # file in cache
-      echo "Found file in cache: ${BUILDPACK_CACHE_DIR}/${name}" >&2
-      echo "${BUILDPACK_CACHE_DIR}/${name}"
+      echo "Found file in cache: ${BUILDPACK_CACHE_DIR}/${filename}" >&2
+      echo "${BUILDPACK_CACHE_DIR}/${filename}"
   else
     # cache disabled or not in cache
-    download_file "${url}" "${name}"
+    download_file "${url}" "${filename}"
   fi
 }
 
@@ -31,7 +36,7 @@ function download_file () {
   local name
   name=${2:-$(basename "${url}")}
 
-  local temp_folder=${BUILDPACK_CACHE_DIR:-$(mktemp -u)}
+  local temp_folder=${BUILDPACK_CACHE_DIR:-${TEMP_DIR}}
   curl --create-dirs -sSfLo "${temp_folder}/${name}" "${url}"
   echo "${temp_folder}/${name}"
 }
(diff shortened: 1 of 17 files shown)


src/usr/local/buildpack/utils/cache.sh
```diff
@@ -0,0 +1,78 @@
...
+# will attempt to download the file at the given url and stores it in the cache.
+# If this file already exists, will return the cached version
+# The cache will only used if BUILDPACK_CACHE_DIR is set
+# First argument is the url, second one is the filename (optional)
+function get_from_url () {
+  local url=${1}
+  check url true
+
+  local name
+  name=${2:-$(basename "${url}")}
```
viceice, line 12: Thisd works for urls? 😏

src/usr/local/buildpack/utils/cache.sh
```diff
@@ -0,0 +1,78 @@
...
+
+  if [ -n "${BUILDPACK_CACHE_DIR}" ] && [ -e "${BUILDPACK_CACHE_DIR}/${name}" ]; then
```
viceice, line 14: We need to make sure the files are unique, can we use `TOOL_NAME` and `TOOL_VERSION` to create some implicit folders?

src/usr/local/buildpack/utils/cache.sh
```diff
@@ -0,0 +1,78 @@
...
+# If the cache is not enabled it will download it to a temp folder
+# The second argument will be the filename if given
+function download_file () {
+  local url=${1}
+  check url true
+
+  local name
+  name=${2:-$(basename "${url}")}
+
+  local temp_folder=${BUILDPACK_CACHE_DIR:-$(mktemp -u)}
```
viceice, line 34: We need to cleanup this when not using cache, as we don't want to have the downloads for prebuild images (eg python or node)

viceice (commented):
Looks good so far

src/usr/local/buildpack/utils/cache.sh
Chumper: yes, urls are not that different than a file path and for our use case this is more than enough

src/usr/local/buildpack/utils/cache.sh
Chumper: I would like not to rely on the tool name and the tool version in this functions. 
Even if we agree that this is currently the only use case we should find a different solution.

The filename is not unique enough for you, so what about if we add the hash of the url to the name?
I would expect the full url to be unique enough without introducing the dependency on the tool name and version.

src/usr/local/buildpack/utils/cache.sh
Chumper: We do, my initial thought was to return the path and the user is responsible for cleanup.
We can introduce a general cache dir that we do not need to clean up like `/opt/buildpack/cache/` for downloaded tools when the cache is disabled. How does that sound?

src/usr/local/buildpack/utils/cache.sh
viceice: hash would be ok for me, as filename isn't unique enough.

src/usr/local/buildpack/utils/cache.sh
viceice: Currently we use `/tmp` which is cleaned up by `install-tool` and others. So can we use that as cache when no explicit caching is enabled?

Chumper: `/tmp` is fine with me for now. 
We still need to have the discussion if we want to move the install folder alltogether but that can be done at a later stage.

viceice (approved):

PR #300, opened 2022-02-13
feat: migrate Helm to a v2 tool

Chumper:
Refactors the first tool to be v2 compatible.


Chumper:
@viceice @rarkins I would like both of your feedback on this.
Improvements that this refactoring contains:

* tars are cached (if the cached is enabled)
* installing only happens once
* when the tool is already installed, it will be linked only
* time for installation is shown
* folder structure is unified for root and non root users

Chumper:
This is a path issue, I will refactor that first

viceice (commented):
Looks good so far

src/usr/local/buildpack/tools/v2/helm.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+function check_tool_requirements () {
```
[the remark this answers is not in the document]
Chumper: we can do that certainly.
In a new PR I can provide a default with the current code and for some tools like `npm` it can override the function to reside inside the `node` folder

Chumper:
@viceice I had to hardcode the `PATH` in the dockerfile. I would have preferred it to be set dynamically with the ìnstall-buildpack` method, but then the file at `/usr/local/etc/env` would not getting parsed for plain `sh` commands. Not sure if you have a solution for that but you did it the same way for the `/home/user/bin` path

src/usr/local/buildpack/tools/v2/helm.sh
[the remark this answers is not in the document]
Chumper: Sorry, I thought we are talking about the `check_tool_installed` function. But it is valid in both cases. I provide a PR for sensitive defaults.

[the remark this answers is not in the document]
Chumper: Done that in #302

Dockerfile
```diff
@@ -12,7 +12,7 @@ LABEL maintainer="Rhys Arkins <rhys@arkins.net>" \
   org.opencontainers.image.source="https://github.com/containerbase/buildpack"
 
 #  autoloading buildpack env
-ENV BASH_ENV=/usr/local/etc/env ENV=/usr/local/etc/env PATH=/home/user/bin:$PATH
+ENV BASH_ENV=/usr/local/etc/env ENV=/usr/local/etc/env PATH=/opt/buildpack/bin:/home/user/bin:$PATH
```
viceice, line 15: @rarkins should we place the symlinks and shell wrapper to `/use/local/bin`, so we don't have to change the `PATH` ?

Chumper: I would prefer not to do that, so that we have a clean separation.
This would be a breaking change but we can do that together with all tools for v3.

viceice: I like to use an existing path folder, as we had and have trouble with it.

Some container runtimes / ci tools override path, so our buildpack tools won't be found.
I've seen this especially on Jenkins.

@rarkins WDYT?

Chumper: So we will set the default bin path to `/usr/local/bin`?
I can make the changes for all v2 functions.

rarkins: I would very much like to use symlinks and not be dynamically switching path

Chumper: 👍 I will change the bin path from `/opt/buildpack/bin` to `/usr/local/bin` in a separate PR

viceice: Ok, we should merge that before this.

rarkins: BTW I'm not wedded to `/usr/local/bin`.. it can be another directory like `/opt/buildpack/bin`. For me the most important part is that the `PATH` doesn't need to keep switching

viceice:
```suggestion
ENV BASH_ENV=/usr/local/etc/env ENV=/usr/local/etc/env PATH=/home/user/bin:$PATH
```

test/Dockerfile.bionic
```diff
@@ -9,7 +9,7 @@ LABEL maintainer="Rhys Arkins <rhys@arkins.net>" \
   org.opencontainers.image.source="https://github.com/containerbase/buildpack"
 
 #  autoloading buildpack env
-ENV BASH_ENV=/usr/local/etc/env ENV=/usr/local/etc/env PATH=/home/$USER_NAME/bin:$PATH
+ENV BASH_ENV=/usr/local/etc/env ENV=/usr/local/etc/env PATH=/opt/buildpack/bin:/home/$USER_NAME/bin:$PATH
```
viceice, line 12:
```suggestion
ENV BASH_ENV=/usr/local/etc/env ENV=/usr/local/etc/env PATH=/home/$USER_NAME/bin:$PATH
```

viceice:
it now needs bats test fixes 🙃

Chumper:
Indeed, but that is already on my local branch

Chumper:
@viceice We might need a new regex for the version checking: https://github.com/containerbase/buildpack/pull/300/files#diff-006c243ef9f027ee0c3b4abd457b5763a6805403a01ddd12abb16b1d1d13f41eR67
Maybe we should extract this in vars and use them in tests

src/usr/local/buildpack/tools/v2/helm.sh
```diff
@@ -0,0 +1,25 @@
...
+
+function link_tool () {
+  local versioned_tool_path
+  versioned_tool_path=$(find_versioned_tool_path)
+
+  link_wrapper "${TOOL_NAME}" "${versioned_tool_path}/bin"
+  helm version
+}
+
+function prepare_tool() {
```
viceice, line 22: Wasn't this function optional? 🤔

viceice (commented):
Otherwise LGTM

viceice:
> @viceice We might need a new regex for the version checking: https://github.com/containerbase/buildpack/pull/300/files#diff-006c243ef9f027ee0c3b4abd457b5763a6805403a01ddd12abb16b1d1d13f41eR67 Maybe we should extract this in vars and use them in tests

Sure, i think I do this in java test already.

src/usr/local/buildpack/tools/v2/helm.sh
Chumper: It is, you have seen nothing :)

viceice (approved):
Do you like to change the test to use env vars in this PR or a new one?

src/usr/local/buildpack/tools/v2/helm.sh
viceice: 😅😴

Chumper:
> Do you like to change the test to use env vars in this PR or a new one?

Will do it in this one, otherwise we never do it

Chumper:
> > Do you like to change the test to use env vars in this PR or a new one?
> 
> Will do it in this one, otherwise we never do it

Done, we also need to do it for the bats test, but for that I will create a ticket

viceice (approved):

Chumper:
I removed the Renovate versioning, you might need to approve one more time

viceice (approved):

PR #305, opened 2022-02-14
fix: create tool folders with 775 for openshift

### LINKED ISSUE #304: User tool folders have wrong rights
If installing a tool like npm as user, the npm folder has 755 but needs 775 as rights, so the root group can add tools when running with openshift.

- Ref whitesource/renovate-on-prem#193

Chumper:
Closes #304
Closes https://github.com/whitesource/renovate-on-prem/issues/193 after updating.

### DIFF
--- a/src/usr/local/buildpack/utils/filesystem.sh
+++ b/src/usr/local/buildpack/utils/filesystem.sh
@@ -33,8 +33,7 @@ function create_versioned_tool_path () {
 
   # shellcheck disable=SC2174
   mkdir -p -m "${umask}" "${install_dir}/${TOOL_NAME}"
-  # shellcheck disable=SC2174
-  mkdir -p -m "${umask}" "${install_dir}/${TOOL_NAME}/${TOOL_VERSION}"
+  mkdir "${install_dir}/${TOOL_NAME}/${TOOL_VERSION}"
   echo "${install_dir}/${TOOL_NAME}/${TOOL_VERSION}"
 }
 
--- a/src/usr/local/buildpack/utils/filesystem.sh
+++ b/src/usr/local/buildpack/utils/filesystem.sh
@@ -25,8 +25,14 @@ function find_versioned_tool_path () {
 
 function create_versioned_tool_path () {
   install_dir=$(get_install_dir)
+
+  local umask=775
+  if [ "$(is_root)" -eq 0 ]; then
+    umask=755
+  fi
+
   # shellcheck disable=SC2174
-  mkdir -p -m 775 "${install_dir}/${TOOL_NAME}"
+  mkdir -p -m "${umask}" "${install_dir}/${TOOL_NAME}"
   mkdir -p "${install_dir}/${TOOL_NAME}/${TOOL_VERSION}"
   echo "${install_dir}/${TOOL_NAME}/${TOOL_VERSION}"
 }
--- a/src/usr/local/buildpack/utils/filesystem.sh
+++ b/src/usr/local/buildpack/utils/filesystem.sh
@@ -32,8 +32,9 @@ function create_versioned_tool_path () {
   fi
 
   # shellcheck disable=SC2174
-  mkdir -p -m "${umask}" "${install_dir}/${TOOL_NAME}"
-  mkdir -p "${install_dir}/${TOOL_NAME}/${TOOL_VERSION}"
+  mkdir -p -m 775 "${install_dir}/${TOOL_NAME}"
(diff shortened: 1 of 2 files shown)


Chumper:
@viceice Let me know if root created tool folders should keep 755. 
In that case I can add a line to make a switch

src/usr/local/buildpack/utils/filesystem.sh
```diff
@@ -25,6 +25,8 @@ function find_versioned_tool_path () {
 
 function create_versioned_tool_path () {
   install_dir=$(get_install_dir)
+  # shellcheck disable=SC2174
+  mkdir -p -m 775 "${install_dir}/${TOOL_NAME}"
   mkdir -p "${install_dir}/${TOOL_NAME}/${TOOL_VERSION}"
```
viceice, line 30: This needs to be 755 when not running as root too.

Chumper: We can change the global umasks to `755` and `775` and use the refactored methods.
How important is `read+execute` for `other` for you?

If not at all, then we can use the current umasks and simply replace the function.

viceice: I think 770 and 750 are ok.

Chumper: I pushed with 775 and 755 first. Lets keep the legacy install compatible and us 770 and 750 for the new tools going forward.

src/usr/local/buildpack/utils/filesystem.sh
```diff
@@ -25,6 +25,14 @@ function find_versioned_tool_path () {
...
   install_dir=$(get_install_dir)
+
+  local umask=775
+  if [ "$(is_root)" -eq 0 ]; then
+    umask=755
+  fi
+
+  # shellcheck disable=SC2174
+  mkdir -p -m "${umask}" "${install_dir}/${TOOL_NAME}"
   mkdir -p "${install_dir}/${TOOL_NAME}/${TOOL_VERSION}"
```
viceice, line 36:
```suggestion
  mkdir -p -m 775 "${install_dir}/${TOOL_NAME}"
   mkdir -p -m "${umask}" "${install_dir}/${TOOL_NAME}/${TOOL_VERSION}"
```
I think this is the right way to go.

viceice: Ups, that wasn't intended. My intention was to have the version folder 755 on user install.

Misunderstanding, sorry.

src/usr/local/buildpack/utils/filesystem.sh
```diff
@@ -25,7 +25,16 @@ function find_versioned_tool_path () {
...
   install_dir=$(get_install_dir)
-  mkdir -p "${install_dir}/${TOOL_NAME}/${TOOL_VERSION}"
+
+  local umask=775
+  if [ "$(is_root)" -eq 0 ]; then
+    umask=755
+  fi
+
+  # shellcheck disable=SC2174
+  mkdir -p -m 775 "${install_dir}/${TOOL_NAME}"
```
viceice, line 35:
```suggestion
  mkdir -p -m "${umask}" "${install_dir}/${TOOL_NAME}"
```
So it should like this? This function is for V1 tools only?

test/bash/filesystem.bats
```diff
@@ -52,6 +54,8 @@ teardown() {
 
     assert_output "${TEST_ROOT_DIR}/root/foo/1.2.3"
     assert [ -d "${TEST_ROOT_DIR}/root/foo/1.2.3" ]
+    assert [ $(stat --format '%a' "${TEST_ROOT_DIR}/root/foo") -eq 775 ]
+    assert [ $(stat --format '%a' "${TEST_ROOT_DIR}/root/foo/1.2.3") -eq 755 ]
```
viceice, line 58:
```suggestion
    assert [ $(stat --format '%a' "${TEST_ROOT_DIR}/root/foo") -eq 755 ]
    assert [ $(stat --format '%a' "${TEST_ROOT_DIR}/root/foo/1.2.3") -eq 755 ]
```

src/usr/local/buildpack/utils/filesystem.sh
Chumper: It is for v1  tools only, but for both - root and user - paths.
Yes, setting the umask in the first `mkdir` only would also be my suggestion. The version should be kept at 755

viceice (approved):

PR #307, opened 2022-02-15
fix: legacy install of npm and yarn

Chumper:
Should fix legacy install of npm and yarn


viceice:
"/tmp/empty-cache":  looks like we need to create this with 775 too

src/usr/local/buildpack/tools/yarn.sh
```diff
@@ -7,11 +7,7 @@ check_command node
...
 
 if [[ -z "${tool_path}" ]]; then
-  INSTALL_DIR=$(get_install_dir)
-  base_path=${INSTALL_DIR}/${TOOL_NAME}
-  tool_path=${base_path}/${TOOL_VERSION}
-
-  mkdir -p "${tool_path}"
+  tool_path="$(create_versioned_tool_path)"
 
   NPM_CONFIG_PREFIX=$tool_path npm install --cache /tmp/empty-cache -g "${TOOL_NAME}@${TOOL_VERSION}"
```
viceice, line 12: We need to clean cache. Looks like it's not removed from install-tool, so next install command with different user fails.

Maybe better use a random path?

viceice (changes requested):

src/usr/local/buildpack/tools/yarn.sh
Chumper: Seems we need to clean this path `/home/user/.npm/_cacache`

Chumper: Seems we need to do a `npm cache clean --force` previously to the install

Chumper:
@viceice I am not sure how reliable these tests are.
I would expect the user to be the same for a single image and not being changed during runtime.
In that case the test to switch to another user for openshift seems wrong to me.

viceice:
The user will change on runtime on openshift.
Openshift will use a arbitrary user id to run the container.

But sure, it then won't run any command as our image user. Maybe we need to use `USER 1005:0` in tests to force root group?

viceice:
I'm also unsure, why npm will access `~/.npm` 🤷‍♂️

src/usr/local/buildpack/tools/yarn.sh
viceice: I thought when using the cache param it will use only that path, but npm seems to still use home dir.

viceice: Looks like a npm bug

- https://github.com/npm/cli/issues/3350

Chumper:
Ok, let me just apply 775 for the folders

viceice:
We need to be sure it now works for openshift.

viceice:
Maybe add a new from line to docker tests and only use 1005 there, so we don't mix install method's

viceice (approved):

PR #322, opened 2022-03-05
fix: use /usr/local/bin as bin path

Chumper:
Changes the bin path from `/opt/buildpack/bin` to `/usr/local/bin` as discussed in #300

### DIFF
--- a/src/usr/local/buildpack/utils/filesystem.sh
+++ b/src/usr/local/buildpack/utils/filesystem.sh
@@ -54,7 +54,7 @@ function setup_directories () {
   mkdir -p -m 770 "$(get_version_path)"
   # contains the wrapper and symlinks for the tools
   # shellcheck disable=SC2174
-  mkdir -p -m 755 "$(get_bin_path)"
+  mkdir -p -m 775 "$(get_bin_path)"
 }
 
 # Creates the given folder path with root and user umask depending on the caller
--- a/src/usr/local/buildpack/utils/filesystem.sh
+++ b/src/usr/local/buildpack/utils/filesystem.sh
@@ -55,6 +55,12 @@ function setup_directories () {
   # contains the wrapper and symlinks for the tools
   # shellcheck disable=SC2174
   mkdir -p -m 775 "$(get_bin_path)"
+
+  # if the bin path exists and does not have 775, force it
+  if [ "$(stat --format '%a' "$(get_bin_path)")" -ne 775 ]; then
+    echo "Forcing 775 on '$(get_bin_path)' ..."
+    chmod 775 "$(get_bin_path)"
+  fi
 }
 
 # Creates the given folder path with root and user umask depending on the caller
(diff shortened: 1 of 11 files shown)


src/usr/local/buildpack/utils/filesystem.sh
```diff
@@ -44,13 +44,17 @@ function setup_directories () {
...
+  # shellcheck disable=SC2174
+  mkdir -p -m 770 "$(get_env_path)"
   # contains the latest version of the tools
-  mkdir -m 770 "$(get_version_path)"
+  # shellcheck disable=SC2174
+  mkdir -p -m 770 "$(get_version_path)"
   # contains the wrapper and symlinks for the tools
-  mkdir -m 770 "$(get_bin_path)"
+  # shellcheck disable=SC2174
+  mkdir -p -m 755 "$(get_bin_path)"
```
viceice, line 57: Does it change permission when directory exist? As `/usr/local/bin` already exist.

viceice (commented):
Otherwise LGTM

viceice:
Merging this as fix later, otherwise we don't get a new build / release.

src/usr/local/buildpack/utils/filesystem.sh
Chumper: It does not, the test failed before when I tested for `770`. So the `-p` checks if the directory exist and if it does simply returns.

viceice: Ok, so we need to make sure it's writable by root group, so we need `775` on it.

src/usr/local/buildpack/utils/filesystem.sh
```diff
@@ -44,13 +44,17 @@ function setup_directories () {
...
+  # shellcheck disable=SC2174
+  mkdir -p -m 770 "$(get_env_path)"
   # contains the latest version of the tools
-  mkdir -m 770 "$(get_version_path)"
+  # shellcheck disable=SC2174
+  mkdir -p -m 770 "$(get_version_path)"
   # contains the wrapper and symlinks for the tools
-  mkdir -m 770 "$(get_bin_path)"
+  # shellcheck disable=SC2174
+  mkdir -p -m 775 "$(get_bin_path)"
```
viceice, line 57: So this only works on tests, as `usr/local/bin` already exist and still has 755?

Chumper: yes, we need to assume that the bin folder in the image already has `775` or we need to make sure to have a docker (=integration) test for it.

So in the end we may need to set up the folder in the `install-buildpack`

Chumper: Let me add an integration test for it and add it to the `install-buildpack` method

viceice: Ok, let's do it in `install-buildpack`

Chumper: I added an integration test, but locally it fails. The bin path is still `755`.
If it fails in the actions as well, then somewhere this is changed back.

viceice (approved):

PR #325, opened 2022-03-05
feat: move all linking targets to /usr/local/bin

Chumper:
Moves the linking path to `/usr/local/bin`

### DIFF
--- a/test/bash/linking.bats
+++ b/test/bash/linking.bats
@@ -17,6 +17,8 @@ setup() {
     USER_HOME="${TEST_ROOT_DIR}/user"
     ENV_FILE="${TEST_ROOT_DIR}/env"
 
+    setup_directories
+
     # set default test user
     TEST_ROOT_USER=1000
 }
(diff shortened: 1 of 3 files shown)


viceice (commented):
needs some test fixes 😏

test/bash/linking.bats
```diff
@@ -39,22 +41,22 @@ teardown() {
 
   run link_wrapper git
   assert_success
-  assert [ -f "${USER_HOME}/bin/git" ]
+  assert [ -f "${BIN_DIR}/git" ]
```
viceice, line 44: So git is just a random test executable?

viceice (approved):

test/bash/linking.bats
Chumper: yes, I can use a more standard executable like ls, cat or sort like that

viceice: Leave as is. 🤗

PR #327, opened 2022-03-06
chore: enable Renovate on helm bats files

### LINKED ISSUE #326: [TESTS] Make Helm version in bats test renovatable
Currently the bats tests for the helm tool are hardcoded and will not be versioned by Renovate.
We should enable all version in the bats tests to be automatically versioned by Renovate.

This probably requires a custom regex manager in the containerbase/.github repository.

Outcome of #300

Chumper:
Makes the helm bats tests versioned by Renovate

Closes #326

### DIFF
--- a/test/bash/tools/helm.bats
+++ b/test/bash/tools/helm.bats
@@ -58,6 +58,7 @@ teardown_file () {
 
 @test "helm: check_tool_installed" {
   local TOOL_NAME=helm
+  local TOOL_VERSION
 
   # renovate: datasource=github-releases lookupName=helm/helm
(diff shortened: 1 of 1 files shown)


viceice (commented):
No longer defined as local?

Chumper:
> No longer defined as local?

I would like to keep local but I haven't looked into the regex expression if that is supported.

viceice:
Simply define as empty local and assign after? 🤗

Chumper:
Some solutions are too simple. Thanks @viceice

viceice (approved):

PR #335, opened 2022-03-11
fix: missing root group for user

### LINKED ISSUE #334: Cannot run Helm with non-root group anymore
The image runs as `1000:0` (`ubuntu:root`) by default, I run it as `1000:1000` (`ubuntu:ubuntu`), and get errors:

```bash
/bin/sh: 1: helm: Permission denied
```

It appears to come from the content of the `/opt/buildpack` directory not being readable for others:

```bash
$ ls -l /opt/buildpack/
total 12
drwxrwx--- 2 root root 4096 Mar  7 21:33 env.d
drwxrwx--- 1 root root 4096 Mar 11 17:17 tools
drwxrwx--- 1 root root 4096 Mar 11 17:17 versions
```

It only seems to be used by Helm in the downstream [renovatebot/docker-renovate-full](https://github.com/renovatebot/docker-renovate-full) image:

```bash
/opt/buildpack/
├── env.d
├── tools
│   └── helm
│       └── 3.8.1
│           └── bin
│               ├── LICENSE
│               ├── README.md
│               └── helm
└── versions
    └── helm
```

The desired permissions would be `775` for directories and executables.

Not sure where that could be changed, around here maybe?:

https://github.com/containerbase/buildpack/blob/566925fdc9d5673693810fc3d67c1aebd9e25c9d/src/usr/local/buildpack/utils/constants.sh#L14-L17

Could have been caused by PR #300

viceice:
Add buildpack user explicit to root group.

- fixes #334

### DIFF
--- a/test/helm/Dockerfile
+++ b/test/helm/Dockerfile
@@ -6,8 +6,6 @@ RUN touch /.dummy
 # renovate: datasource=github-releases lookupName=helm/helm
 ARG HELM_VERSION=3.8.0
 
-USER 1000
-
 #--------------------------------------
 # install helm as root
 #--------------------------------------
@@ -49,6 +47,8 @@ RUN helm version | grep "${HELM_VERSION}"
 #--------------------------------------
 FROM build as testc
 
+USER 1000
+
 RUN install-tool helm "v${HELM_VERSION}"
 
 USER 1111
@@ -67,6 +67,23 @@ SHELL [ "/bin/sh", "-c" ]
 RUN helm version | grep "3.7.0"
 
 #--------------------------------------
+# install helm user 1000:1000
+#--------------------------------------
+FROM build as testd
+
+USER 1000:1000
+
+RUN install-tool helm "v${HELM_VERSION}"
+
+RUN set -ex; \
+  helm repo add stable https://charts.helm.sh/stable; \
+  helm repo update
+
+SHELL [ "/bin/sh", "-c" ]
+
+RUN helm version | grep "${HELM_VERSION}"
+
+#--------------------------------------
 # final
 #--------------------------------------
 FROM build
@@ -74,3 +91,4 @@ FROM build
 COPY --from=testa /.dummy /.dummy
 COPY --from=testb /.dummy /.dummy
 COPY --from=testc /.dummy /.dummy
+COPY --from=testd /.dummy /.dummy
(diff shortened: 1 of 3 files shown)


viceice:
@Chumper @rarkins  this doesn't fix the issue, even if the user is now a member of root group. 😕 Should we require the root group membership?

viceice:
Maybe docker is blocking the other group memberships?

Chumper:
Having the root group is easier, but we can also tweak it that the root group is not needed. 
But that would mean opening `/usr/local/bin` for everyone to be writable.
Not sure if that is expected.

The alternative is to require root group during installation but allow executtion for others as well, so as said in the ticket making the permission `755`

test/helm/Dockerfile
```diff
@@ -66,6 +68,23 @@ SHELL [ "/bin/sh", "-c" ]
 
 RUN helm version | grep "3.7.0"
 
+#--------------------------------------
+# install helm user 1000:1000
+#--------------------------------------
+FROM build as testd
+
+USER 1000:1000
```
Chumper, line 76: This should not work and we should not support it.

Either we should install as `1000:0` and execute as `1000:1000`

viceice:
> Having the root group is easier, but we can also tweak it that the root group is not needed. But that would mean opening `/usr/local/bin` for everyone to be writable. Not sure if that is expected.
> 
> The alternative is to require root group during installation but allow executtion for others as well, so as said in the ticket making the permission `755`

We need the root group, as openshift is running all container as `USER <random-uild>:0` 😕

Chumper:
Yeah, I think expecting the user to be a member of the root group during installation is fine and we should keep that.
Expecting the user to be a member of the root group during execution is something we can change. 
So that even a non-root group member can execute the tools.
Does that make sense?

test/helm/Dockerfile
viceice: sure, was for testing only. will revert. for fast fix we should make the new paths reable and executable by others. that should hopefully fix at leat running the full renovate image with `1000:1000`?

@rarkins WDYT? Should we support running renovate image as `1000:1000`? That won't support `binarySource=install` as that requires root group to write to `/opt/buildpack`. I don't like to make all those folders writable by all.

Chumper: Yes, `775` for non-root install and `755` for root install should be fine.
So the change you made on the `setup_directories` is fine.
Additionally you need to change the usmask here: https://github.com/containerbase/buildpack/blob/main/src/usr/local/buildpack/utils/constants.sh#L15-L17

And adapt the tests :)

rarkins:
We need to assume that Openshift will run us as something like 4746474:0

PR #341, opened 2022-03-16
fix: shell wrapper has 775 permissions for overwrites

### LINKED ISSUE #338: User installs should be able to overwrite tool symlinks created by root installs
When a root installs a tools the symlink in `/usr/local/bin` will have the mode `755` set.
This prevents user installs to overwrite symlinks when a user installs a tool.

We should change this to be `775` on created symlinks,

Chumper:
* Sets the correct path for shell wrappers
* Adds test for the shell wrapper
* sets the permissions to `775` for the shell wrappers

Closes #338

### DIFF
--- a/src/usr/local/buildpack/utils/linking.sh
+++ b/src/usr/local/buildpack/utils/linking.sh
@@ -3,7 +3,7 @@
 # use this if custom env is required, creates a shell wrapper to /usr/local/bin
 function shell_wrapper () {
   local FILE
-  FILE="$(get_bin_path)/bin/${1}"
+  FILE="$(get_bin_path)/${1}"
   check_command "$1"
   cat > "$FILE" <<- EOM
 #!/bin/bash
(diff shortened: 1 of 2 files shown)


src/usr/local/buildpack/utils/linking.sh
```diff
@@ -3,7 +3,7 @@
 # use this if custom env is required, creates a shell wrapper to /usr/local/bin
 function shell_wrapper () {
   local FILE
-  FILE="$(get_bin_path)/bin/${1}"
+  FILE="$(get_bin_path)/${1}"
```
viceice, line 6: Why this change?

viceice (commented):
Otherwise LGTM

src/usr/local/buildpack/utils/linking.sh
Chumper: because the `bin` in the path is already returned by `get_bin_path`
I checked the usages, apparently nothing uses this function yet, that is why it did not broke.

Chumper: I introduced the error in this PR: https://github.com/containerbase/buildpack/pull/275/files#diff-5a9264b9dd9a3e58f84880c628edc05ab79a00476745f5836ba8d2123254f26eL6

PR #343, opened 2022-03-17
fix: Revert "fix: use new poetry install url"

viceice:
Reverts containerbase/buildpack#342


viceice:
@Chumper flaky script download. 😕

PR #345, opened 2022-03-18
feat: abort on missing file url

Chumper:
Allows the cache to return non-existing file urls

This required for tools like python that will compile the binaries when no pre-build binaries are found

### DIFF
--- a/src/usr/local/buildpack/utils/cache.sh
+++ b/src/usr/local/buildpack/utils/cache.sh
@@ -40,9 +40,8 @@ function download_file () {
   name=${2:-$(basename "${url}")}
 
   local temp_folder=${BUILDPACK_CACHE_DIR:-${TEMP_DIR}}
-  if curl --create-dirs -sSfLo "${temp_folder}/${name}" "${url}" >/dev/null 2>&1 ; then
-    echo "${temp_folder}/${name}"
-  fi
+  curl --create-dirs -sSfLo "${temp_folder}/${name}" "${url}" >/dev/null 2>&1 || exit 1
+  echo "${temp_folder}/${name}"
 }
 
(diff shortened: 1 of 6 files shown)


viceice:
We will remove the compile on the fly feature. It's no longer needed, but we should do a major bump for removal.

viceice (commented):
As said, i don't think we need this. Makes it more complicated.

Chumper:
Ok, so when the file cannot be donwloaded the cache path should be returned regardless?

viceice:
Sure, if it exists. So it should fail if cache is missing and download fails.

Chumper:
Ok, in that case we should merge this PR. It will return an empty path if the file does not exist in the cache or cannot be downloaded.

Or would you rather have it exit with a 1 status code?

viceice:
Let's exit with error, so we don't need an extra check in tool code.

src/usr/local/buildpack/utils/cache.sh
```diff
@@ -37,8 +40,9 @@ function download_file () {
   name=${2:-$(basename "${url}")}
 
   local temp_folder=${BUILDPACK_CACHE_DIR:-${TEMP_DIR}}
-  curl --create-dirs -sSfLo "${temp_folder}/${name}" "${url}"
-  echo "${temp_folder}/${name}"
+  if curl --create-dirs -sSfLo "${temp_folder}/${name}" "${url}" >/dev/null 2>&1 ; then
+    echo "${temp_folder}/${name}"
+  fi
```
viceice, line 45: Didn't we aggreed download should fail all? 🤔

Chumper: we did, but I haven't changed the line yet 🙈

PR #346, opened 2022-03-18
feat: replace symlink with simple shell wrapper

Chumper:
Replaces the symlink for tools with a simple shell wrapper

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM containerbase/node:16.14.1@sha256:39877f7e8b4d711680ef77344c57f413fedc9644db463338ed9f197fb1bb6e5b
+FROM containerbase/node:16.14.2@sha256:0584f9ac394c568e96576027ac6b2ecb96ece16df64ebfb9f6843033c1a1b5aa
 
 # renovate: datasource=npm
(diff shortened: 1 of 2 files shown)


Chumper:
Closes #338

viceice (commented):
Node tests are failing. 🤔

Chumper:
yes, but i don't understand why... I need to dig deeper to understand the problem

Chumper:
I am not able to find the error. I suspect due to the shell script, paths have been changed but even a diff of the working and failing run does not bring me closer to a solution.

Chumper:
Apparently with `ln` npm gets installed in `/home/user/npm/8.0.0`, with the shell script in `/home/user/.npm-global/bin/npm`

viceice:
We should not change `link_wrapper`, instead we should move all tools to `shell_wrapper` and v2 tools.

florianklumb:
Any verdict on when this will be deployed (into Docker image)?

Chumper:
We opted for refactoring all tool afaik instead of replacing this for all tools at once.

viceice:
@florianklumb So you have issues? PLease open a discussion, so we can checkout your usecase. And then we can open an issue if there is a feature request or bug

PR #354, opened 2022-04-05
feat: use shell_wrapper for all v2 tools

Chumper:
uses the `shell_wrapper` for all current v2 tools


Chumper:
yes, I will add that to the download function and refactor the node tool to use it.

PR #358, opened 2022-04-16
chore: use global cache for tool tests if possible

Chumper:


test/bash/tools/flux.bats
```diff
@@ -40,10 +44,6 @@ teardown() {
     rm -rf "${TEST_ROOT_DIR}"
 }
 
-teardown_file () {
-  rm -rf "${BUILDPACK_CACHE_DIR}"
```
viceice, line 44: shouldn't we cleanup cache dir, if we use a temporary folder? wouldn't it otherwise create new folders on every run on local dev machine?

viceice (commented):
otherwise LGTM

test/bash/tools/flux.bats
Chumper: indeed, will add that

test/bash/tools/flux.bats
```diff
@@ -1,7 +1,11 @@
 setup_file () {
-  # set up the cache
-  BUILDPACK_CACHE_DIR="$(mktemp -u)"
-  mkdir -m 777 -p "${BUILDPACK_CACHE_DIR}"
+  # use global cache
+  mkdir -m 777 -p /tmp/renovate/bats || true
+  if [ -e /tmp/renovate/bats ]; then
```
viceice, line 4: wouldn't this always be true for tests?

PR #402, opened 2022-06-11
chore: allow bats tests to use a global cache dir

Chumper:
Example usage for local testing:
```bash
export TEST_CACHE_DIR=/tmp/renovate/bats
yarn bats
```
Reduces the bats testing from 16 to 6 seconds when the cache if filled.


viceice (approved):

viceice:
not sure what's going on but approved the PR.

so hopefully a new run later will fix and merge this PR

PR #403, opened 2022-06-12
feat: add node as v2 tool

Chumper:
Converts the following tools:
* node
* npm
* yarn
* pnpm


Chumper:
@viceice Currently working on the npm conversion. WIll do both together in this one, so that I can refactor some code to make it more composable.

viceice (commented):
sure, go ahead.

src/usr/local/buildpack/utils/node.sh
```diff
@@ -7,18 +7,49 @@ function npm_init () {
...
+    "$versioned_tool_path/bin/npm" explore npm --global --prefix "$versioned_tool_path" -- npm install node-gyp@latest --no-audit --cache "${NPM_CONFIG_CACHE}" 2>&1
   fi
 }
 
 function npm_clean () {
   # Clean npm stuff
   rm -rf "$HOME/.cache" "${NPM_CONFIG_CACHE}" "$HOME/.npm/_logs"/*
 }
+
+# Helper function to link to a globally installed node
```
viceice, line 29: shouldn't those functions stay in node tool? as they are only used / allowed for node installer.

viceice (commented):
can you convert pnpm and yarn too? they belong to node too.

src/usr/local/buildpack/utils/node.sh
```diff
@@ -7,14 +7,17 @@ function npm_init () {
...
 function npm_install () {
-  local tool_path=
-  tool_path="$(create_versioned_tool_path)"
+  local versioned_tool_path
+  versioned_tool_path="$(create_versioned_tool_path)"
 
-  npm install "${TOOL_NAME}@${TOOL_VERSION}" --global --no-audit --prefix "$tool_path" --cache "${NPM_CONFIG_CACHE}" 2>&1
+  # get semver => MAJOR, MINOR, PATCH
+  check_semver "${TOOL_VERSION}"
+
```
viceice, line 15:
```suggestion
```
didn't we agreed this isn't required?

src/usr/local/buildpack/tools/v2/yarn.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+function check_tool_requirements () {
+  check_command node
+  check_semver "$TOOL_VERSION"
+
+  if [[ ! "${MAJOR}" || ! "${MINOR}" || ! "${PATCH}" ]]; then
+    echo Invalid version: "${TOOL_VERSION}"
+    exit 1
+  fi
```
viceice, line 10: can we create a helper to reduce duplicate code?

Chumper: will do

src/usr/local/buildpack/util.sh
```diff
@@ -68,6 +68,22 @@ function check_semver () {
   export PATCH=${BASH_REMATCH[5]}
 }
 
+function validate_semver () {
+  local number_of_parts=${1:-3}
```
viceice, line 72: shouldn't we call `check_semvrr` here? so we've one less code line?

Chumper: yes, we can do that

viceice: better pass version as parameter too, so this function doesn't need to know the env name

viceice (approved):

PR #404, opened 2022-06-12
feat: convert python to a v2 tool

Chumper:


src/usr/local/buildpack/tools/v2/python.sh
```diff
@@ -0,0 +1,101 @@
...
+
+  # get semver -> major, minor, patch
+  check_semver "${TOOL_VERSION}"
+
+  reset_tool_env
+
+  # export python vars
+  export_tool_env PYTHONHOME "${versioned_tool_path}"
+  export_tool_path "${versioned_tool_path}/bin"
+  export_tool_path "${USER_HOME}/.local/bin"
```
Chumper, line 89: probably not needed anymore, because we have one install path

src/usr/local/buildpack/tools/v2/python.sh
```diff
@@ -0,0 +1,101 @@
...
+      sed -i "1 s:.*:#\!${versioned_tool_path}\/bin\/python${MAJOR}:" "${file}"
+      ;;
+    "#!"*"/bin/python${MAJOR}.${MINOR}" )
+      sed -i "1 s:.*:#\!${versioned_tool_path}\/bin\/python${MAJOR}.${MINOR}:" "${file}"
+      ;;
+    esac
+  done < <(find "${versioned_tool_path}/bin" -type f -exec grep -Iq . {} \; -print0)
+}
+
+function prepare_tool() {
```
viceice, line 22: i think we should do a major version bump and drop this. so we no longer need any prepare step.

will do this in a separate PR today before this PR.

src/usr/local/buildpack/tools/v2/python.sh
```diff
@@ -0,0 +1,101 @@
...
+
+  create_folder "${versioned_tool_path}/bin"
+
+  file=$(get_from_url "${url}/${TOOL_VERSION}/python-${TOOL_VERSION}-${version_codename}-${arch}.tar.xz")
+
+  if [[ -f ${file} ]]; then
+    echo 'Using prebuild python'
+    tar -C "${versioned_tool_path}" --strip 1 -xf "${file}"
+  else
+    python-build "$TOOL_VERSION" "${versioned_tool_path}/${TOOL_VERSION}"
```
viceice, line 65: see other comment.

src/usr/local/buildpack/tools/v2/python.sh
viceice: it's required for `pip install ...` support

src/usr/local/buildpack/tools/v2/python.sh
```diff
@@ -0,0 +1,101 @@
...
+    ;
+  if [[ ! -x "$(command -v python-build)" ]]; then
+    git clone https://github.com/pyenv/pyenv.git
+    pushd pyenv/plugins/python-build || exit
+    ./install.sh
+    popd || exit
+    rm -rf pyenv
+  fi
+}
+
```
viceice, line 41:
```suggestion
```

src/usr/local/buildpack/tools/v2/python.sh
viceice:
```suggestion
  else
    echo 'No prebuild python found' >&2
    exit 1
```

src/usr/local/buildpack/tools/v2/python.sh
```diff
@@ -0,0 +1,82 @@
...
+function install_tool () {
+  local versioned_tool_path
+  local file
+  local url
+  local arch
+  local version_codename
+
+  # get semver -> major, minor, patch
+  check_semver "${TOOL_VERSION}"
```
viceice, line 30: Why do we need this again?

Chumper: In this case MAJOR, MINOR and PATCH are unset, beause they get only exported in the requirements check.
If you have a better idea, let me know.
We can also move it one level up, but that might not make sense for all tools

viceice: 🤔 i think those variables are always availabe after  `check_semver` call, because they are explict exported

https://github.com/containerbase/buildpack/blob/00ec26c2c88bcc59c7b0fddeded9da296140f56f/src/usr/local/buildpack/util.sh#L66-L68

Soo i thinnk we don't need to call again

Chumper: I can remove that line, but my feeling is that it won't work then. 
Let me do that

Chumper: Let me check if I did anything wrong

Chumper: I did indeed

viceice (approved):

PR #425, opened 2022-06-22
fix: measure install time and echo on stdout

Chumper:


viceice (approved):
can we also measure legacy installs?

viceice (approved):

PR #431, opened 2022-06-26
feat(v2): add generic cache for tools

Chumper:
Enables tools to use a generic cache to prevent new installs everytime


src/usr/local/buildpack/utils/cache.sh
```diff
@@ -1,5 +1,27 @@
 #!/bin/bash
 
+# This is the cache implementation for the buildpack project.
+#
+# The idea is to cache tools to speed up the installation for the next installs.
+# For this this file supports caching of the following:
```
viceice, line 6:
```suggestion
# For this file supports caching of the following:
```
still sounds wrong 🤔

src/usr/local/buildpack/utils/cache.sh
```diff
@@ -1,5 +1,27 @@
...
+# The structure of the cache is the following:
+# ${BUILDPACK_CACHE_DIR}
+# ├── 3018548ea9a02ddc5ba69b0044ca78bb9e65bfe0
+# │   └── binary.tar.xz
+# ├── 375f537df492a02fd1744a6e0973943e45163a10
+# │   └── folder.tar
+#
+# The cache is disabled by default and will only be activated when BUILDPACK_CACHE_DIR is set.
+# It also supports cleaning up the cache if there is not enough disk space.
+# Cleanup will always happen on the checksum folder, not the individual files.
```
viceice, line 22: `um` ?

src/usr/local/buildpack/utils/cache.sh
```diff
@@ -85,3 +107,82 @@ function get_cache_fill_level () {
...
+  # create folder
+  mkdir -p "${BUILDPACK_CACHE_DIR}/${checksum}"
+
+  # remove file if it exists
+  if [ -e "${cache_path}" ]; then
+    rm "${cache_path}"
+  fi
+
+  # tar folder
+  tar -cf "${cache_path}" -C "${path}" .
```
viceice, line 151: tgz is very slow, we should use xz compression, or better zstd

src/usr/local/buildpack/utils/cache.sh
```diff
@@ -85,3 +107,82 @@ function get_cache_fill_level () {
...
+  fi
+
+  local checksum
+  checksum=$(calculate_checksum "${key}")
+
+  local filename="${checksum}/folder.tar"
+  local cache_path="${BUILDPACK_CACHE_DIR}/${filename}"
+
+  # create folder
+  mkdir -p "${BUILDPACK_CACHE_DIR}/${checksum}"
```
viceice, line 143: we need to set writable flag for root group

src/usr/local/buildpack/utils/cache.sh
Chumper: I am not compressing the files here, I basically just need the folder to be one file, we could use zip or whatever.
I opted for tar without compression.

src/usr/local/buildpack/utils/cache.sh
Chumper: Will use the utils class here.

src/usr/local/buildpack/utils/cache.sh
Chumper: @viceice Do we still need to change this and use compression?

viceice: yes i think so, as the app otherwise is getting out of Disk space more often

Chumper: 👍 will do

viceice: - install `zstd` with apt_install
- compress: `tar -C $path -cf - . | zstd --no-progress -z -T0 --long=30 -o /tmp/$file`
- uncompress: `zstd --no-progress -c -d -T0 --long=30 /tmp/$file | tar -C $path -xpf -`

Using this on my ci to compress node_modules and dotnet packages folder, is pretty fast.
Maybe we can play with parameters to weigth speed over compression.

src/usr/local/buildpack/utils/cache.sh
```diff
@@ -85,3 +112,82 @@ function get_cache_fill_level () {
   check BUILDPACK_CACHE_DIR true
   df --output=pcent "${BUILDPACK_CACHE_DIR}" | tr -dc '0-9'
 }
+
+# Will calculate the checksum for the given string
+function calculate_checksum () {
+  local text=$1
+  echo "${text}" | sha1sum | awk '{print $1}'
```
viceice, line 119: shouldn't we use sha256 instead, sha1 is deprecated and usage is discouraged

viceice (approved):

PR #448, opened 2022-07-16
test: add erlang v25 test

rarkins:
- blocked by #518


viceice (commented):
something seems to be wrong with v25

viceice:
Elixir not yet compatible with erlang v25 https://hexdocs.pm/elixir/1.13.4/compatibility-and-deprecations.html#compatibility-between-elixir-and-erlang-otp

viceice:
elixir v1.14 will support erlang v25 and will probably released nex 2 weeks

Blond11516:
Elixir 1.14.0 has been released with support for Erlang 25. However, the format of assets in the github release has changed and is not longer compatible with the url used by `install-tool`: https://github.com/elixir-lang/elixir/releases/tag/v1.14.0

I suggest downloading releases from Hex instead, as does the [asdf elixir plugin](https://github.com/asdf-vm/asdf-elixir/blob/d08f506b72faddff571b819f586219295bf89d99/bin/install#L151). You can see the full list of versions available on Hex [here](https://repo.hex.pm/builds/elixir/builds.txt)

viceice:
> Elixir 1.14.0 has been released with support for Erlang 25. However, the format of assets in the github release has changed and is no longer compatible with the url used by `install-tool`: https://github.com/elixir-lang/elixir/releases/tag/v1.14.0
> 
> I suggest downloading releases from Hex instead, as does the [asdf elixir plugin](https://github.com/asdf-vm/asdf-elixir/blob/d08f506b72faddff571b819f586219295bf89d99/bin/install#L151). You can see the full list of versions available on Hex [here](https://repo.hex.pm/builds/elixir/builds.txt)

Doesn't work for older releases, so did another workaround

viceice (approved):

Issue #462, opened 2022-07-28
V2 Tools can't be re-linked


viceice:
The `check_semver` is skipped when tool already installed. But some tools (like `python` need the version parts in `link_tool` .

Which causes `No /opt/buildpack/tools/python/3.10.5/bin/python. defined - aborting` see dot after `python`, should be `python3.10` instead.

Found while working on:
- #461

PR #466, opened 2022-07-30
fix: allow linking an already installed version

janrotter:
While installing an already present version of python with `install-tool`, the linking process fails (if it is called, i.e. if there is a mismatch between the current version and the one to be installed) because the `MAJOR` and `MINOR` env variables are not set and the constructed path is invalid:

```bash
# src/usr/local/buildpack/tools/v2/python.sh
function link_tool () {
  ...
  python_shell_wrapper "${TOOL_NAME}${MAJOR}" "${versioned_tool_path}"
  python_shell_wrapper "${TOOL_NAME}${MAJOR}.${MINOR}" "${versioned_tool_path}"
```

These variables are set by the `check_semver` function, that is called by the `check_tool_requirements` in the `install_v2_tool` (`src/usr/local/buildpack/utils/install.sh`), but that invocation is skipped when the tool is already present and only requires linking.

This change adds a suitable test and the `check_semver` call in the `link_tool` definition for python to address this issue.

- fixes https://github.com/containerbase/buildpack/issues/462

### DIFF
--- a/src/usr/local/buildpack/tools/v2/python.sh
+++ b/src/usr/local/buildpack/tools/v2/python.sh
@@ -95,8 +95,6 @@ function install_tool () {
 function link_tool () {
   local versioned_tool_path
   versioned_tool_path=$(find_versioned_tool_path)
-  
-  check_semver "${TOOL_VERSION}"
 
   reset_tool_env
 
--- a/src/usr/local/buildpack/tools/v2/python.sh
+++ b/src/usr/local/buildpack/tools/v2/python.sh
@@ -95,6 +95,8 @@ function install_tool () {
 function link_tool () {
   local versioned_tool_path
   versioned_tool_path=$(find_versioned_tool_path)
+  
+  check_semver "${TOOL_VERSION}"
 
   reset_tool_env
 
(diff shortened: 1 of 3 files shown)


viceice (changes requested):
this only fixes it for python, we need to fix it for all tools

- #462

janrotter:
Thank you @viceice for a quick response.

Would always calling the `check_tool_requirements` in the `install_v2_tool` be a better solution?
```diff
diff --git a/src/usr/local/buildpack/utils/install.sh b/src/usr/local/buildpack/utils/install.sh
index bdae973..adc34da 100644
--- a/src/usr/local/buildpack/utils/install.sh
+++ b/src/usr/local/buildpack/utils/install.sh
@@ -12,9 +12,10 @@ function install_v2_tool () {
   # shellcheck source=/dev/null
   . "${path}"
 
+  check_tool_requirements
+
   if ! check_tool_installed; then
     echo "installing v2 tool ${TOOL_NAME} v${TOOL_VERSION}"
-    check_tool_requirements
     install_tool
   else
     echo "tool ${TOOL_NAME} v${TOOL_VERSION} is already installed"
```

viceice:
yes i think so.

@Chumper WDYT?

Chumper:
Yes, simply add it to the install method, there is no harm in doing that.

viceice (approved):

rarkins (approved):

viceice (approved):

PR #491, opened 2022-08-18
docs(custom-certs): add Java configuration section

viceice:
- #3

### DIFF
--- a/docs/custom-root-ca.md
+++ b/docs/custom-root-ca.md
@@ -46,7 +46,7 @@ docker run --rm -it \
 
 ### Java
 
-For java you need to mount your own certificate store to `/opt/buildpack/ssl/cacerts`.
+For Java you need to mount your own certificate store to `/opt/buildpack/ssl/cacerts`.
 
 ```bash
 docker run --rm -it \
--- a/docs/custom-root-ca.md
+++ b/docs/custom-root-ca.md
@@ -18,7 +18,19 @@ RUN update-ca-certificates
 ENV NODE_EXTRA_CA_CERTS=/usr/local/share/ca-certificates/my-root-ca.crt
 ```
 
-**TODO**: For java based tools we need another option
+### Java
+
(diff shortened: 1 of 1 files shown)


docs/custom-root-ca.md
```diff
@@ -31,3 +43,16 @@ docker run --rm -it \
   -e NODE_EXTRA_CA_CERTS=/my-root-ca.crt \
   containerbase/buildpack bash
 ```
+
+### Java
+
+For java you need to mount your own certificate store to `/opt/buildpack/ssl/cacerts`.
```
viceice, line 49:
```suggestion
For Java you need to mount your own certificate store to `/opt/buildpack/ssl/cacerts`.
```

docs/custom-root-ca.md
```diff
@@ -31,3 +43,16 @@ docker run --rm -it \
   -e NODE_EXTRA_CA_CERTS=/my-root-ca.crt \
   containerbase/buildpack bash
 ```
+
+### Java
+
+For Java you need to mount your own certificate store to `/opt/buildpack/ssl/cacerts`.
```
rarkins, line 49: How is this URL used when it's only mapped at runtime? e.g. will any java we install automatically look for this URL even if it wasn't present at the time of prepare-tool?

viceice: - `/opt/buildpack/ssl` is created by `install-buildpack`
- `/opt/buildpack/ssl/cacerts` is created by `prepare-tool java` if not exists
- when running `install-tool java <version>` as root, it automatically runs `prepare-tool java`

rarkins: So the certs file always exists and is expected by Java to be present? That way when it's overridden at runtime Java is already configure to use it?

viceice: yes, it's symlinked to the normal path
https://github.com/containerbase/buildpack/blob/5aed1e46fb27b0b03a1aaa4fd583546d24e11e4c/src/usr/local/buildpack/utils/java.sh#L74-L78

rarkins (approved):

PR #511, opened 2022-09-02
test: fix poetry tests by removing poetry 1.1.0

rarkins:


rarkins:
Let's unblock the repo and add an issue to fix old poetry

PR #575, opened 2022-11-01
feat!: rename user to ubuntu

### LINKED ISSUE #572: Proposal: user -> ubuntu
Reasons to do this:
- Containerbase only works using Ubuntu images today
- Changing the base image in future would be a large task, definitely a breaking change anyway
- Renovate and Mend images have been using `ubuntu` as the user for so long that changing them to use `user` instead would be much more work

So therefore my proposal is to refactor Containerbase to default to `ubuntu` instead of `user` here: https://github.com/containerbase/base/blob/5354700b5b7dc262783ea45a642dd82e17241a51/src/usr/local/bin/install-buildpack#L8

This would allow us to deprecate the `renovate/*` containerbase (buildpack) images faster.

Any reasons why not to do this?

viceice:
Rename default user from `user` back to `ubuntu`

- closes #572

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ LABEL maintainer="Rhys Arkins <rhys@arkins.net>" \
   org.opencontainers.image.source="https://github.com/containerbase/buildpack"
 
 #  autoloading buildpack env
-ENV BASH_ENV=/usr/local/etc/env ENV=/usr/local/etc/env PATH=/home/user/bin:$PATH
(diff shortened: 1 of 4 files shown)


rarkins (approved):
Surprised to learn we have some hardcoded PATHs though

viceice:
> Surprised to learn we have some hardcoded PATHs though

yes, there the env isn't yet available 🙃

PR #634, opened 2022-12-21
feat(poetry): convert to v2 tool

viceice:
Should we use a venv instead?

~venv would be bigger, as it also copies python inside~


Chumper:
We can try out venv if it is easier to handle. We are basically doing the same in a poor mans way.

viceice:
> We can try out venv if it is easier to handle. We are basically doing the same in a poor mans way.

created #635 for `venv`

src/usr/local/buildpack/utils/python.sh
```diff
@@ -0,0 +1,73 @@
...
+
+function check_tool_requirements () {
+  check_command python
+  check_semver "$TOOL_VERSION" "all"
+}
+
+function get_python_minor_version() {
+  local python_version=$1
+  if [[ ! "${python_version}" =~ ${SEMVER_REGEX} ]]; then
+    echo Ruby is not a semver like version - aborting: "${python_version}"
```
Chumper, line 11: s/Ruby/Python/

PR #644, opened 2022-12-23
feat(node): add renovate installer

viceice:
Adds `renovate` install tool

Should i add `source-map-support` for better typescript stack traces?


rarkins:
source map could be very useful

viceice:
> source map could be very useful

Let's add it to the renovate cli directly instead.

PR #687, opened 2023-01-30
feat!: drop unzip

### LINKED ISSUE #502: remove `unzip` and use  `bsdtar` instead
`unzip` is no longer used and can be removed, but it's maybe used by extended images.

viceice:
- closes #502

```
BREAKING CHANGE: unzip executable is no longer available
```


viceice:
flutter needs unzip

PR #713, opened 2023-02-05
feat: basic arm64 support

### LINKED ISSUE #711: ARM support
All `containerbase` images are only built for amd64, it would be great if they were also build for arm64. This would allow dev containers based on `containerbase` (e.g. renovate) to run faster and more efficiently on ARM-based machines like the newer M1 and M2 Apple Macbooks.

**Tasks**
- [x] `dart` #772
- [x] `erlang` #719
- [x] `flutter` #807 
- [x] `java` #500
- [x] `ruby` containerbase/ruby-prebuild#7
- [x] `php` containerbase/php-prebuild#17 #814
- [x] `python` containerbase/python-prebuild#2
- [x] `swift` #805
- [ ] publish multiarch image (amd64 and arm64)
- [ ] renovatebot/renovate#9229

viceice:
- closes #47
- ref #711


Chumper (approved):
FIne with me.
```
 if [[ "$(uname -p)" = "aarch64" ]]; then
    arch=linux-arm64
  fi
```
There are a lot of duplicates, we should maybe refactor that into a common place

viceice:
> FIne with me.
> ```
>  if [[ "$(uname -p)" = "aarch64" ]]; then
>     arch=linux-arm64
>   fi
> ```
> There are a lot of duplicates, we should maybe refactor that into a common place

yes, will do in a follow up PR

PR #725, opened 2023-02-15
feat(tools): support checksum for downloaded files, enable for nodejs

Chumper:


src/usr/local/buildpack/tools/v2/node.sh
```diff
@@ -40,8 +40,17 @@ function install_tool () {
   local versioned_tool_path
   local npm # temp npm executable
 
+  checksums=$(get_from_url "https://nodejs.org/dist/v${TOOL_VERSION}/SHASUMS256.txt")
+
+  # get checksum from file
+  original_checksum=$(grep "node-v${TOOL_VERSION}-linux-x64.tar.xz" "${checksums}" | cut -d' ' -f1)
```
viceice, line 46: please move the `linux-x64` to a local variable, will need that for arm64 support. checkout the open draft PR.

src/usr/local/buildpack/utils/cache.sh
```diff
@@ -49,28 +58,107 @@ function get_from_url () {
...
+  local expected_checksum
+  expected_checksum=${3}
+
+  local checksum_algo
+  checksum_algo=${4}
+
+  local retry=3
+  while [ "${retry}" -gt 0 ]; do
+    retry=$((retry-1))
+    local temp_folder=${BUILDPACK_CACHE_DIR:-${TEMP_DIR}}
```
viceice, line 86: shouldn't this be outside the loop because it's static?

src/usr/local/buildpack/utils/cache.sh
```diff
@@ -49,28 +58,107 @@ function get_from_url () {
       echo "${BUILDPACK_CACHE_DIR}/${filename}"
```
viceice, line 58: validate here too, so we catch invalid cache

src/usr/local/buildpack/utils/cache.sh
Chumper: It should, will move code around

viceice (approved):

PR #736, opened 2023-02-27
fix(python): install python tools with user umask

Chumper:


test/python/Dockerfile
```diff
@@ -115,6 +115,36 @@ ARG BUILDPACK_DEBUG
...
+RUN install-tool python 3.10.0
+
+# renovate: datasource=pypi
+RUN install-tool poetry 1.3.2
+
+RUN set -ex \
+  && cd c-poetry \
+  && poetry update --lock --no-interaction \
+  && poetry add h3py \
+  ;
```
viceice, line 146: i think we can remove this to speed up tests

test/python/Dockerfile
```diff
@@ -115,6 +115,36 @@ ARG BUILDPACK_DEBUG
...
+ARG BUILDPACK_DEBUG
+
+# renovate: datasource=pypi
+RUN install-tool poetry 1.3.2
+
+RUN set -ex \
+  && cd c-poetry \
+  && poetry update --lock --no-interaction \
+  && poetry add h3py \
+  ;
```
viceice, line 132: see below, not really relevant and slows tests.

viceice (commented):
otherwise LGTM

viceice (approved):

Issue #715, opened 2023-02-07
failed tool installation but still considered installed


korosuke613:
## Summary

When installing a new version of Node.js, the file download fails for some reason. However, an empty directory will remain to store the files for that version. Whether or not a version of Node.js is already installed is determined by the presence or absence of that version's directory. Thus, even if Node.js is not installed, it is assumed to be installed, and subsequent processing will fail due to missing files.

## Details

We are [self-hosting Renovate](https://github.com/renovatebot/docker-renovate), which depends on ghcr.io/containerbase/buildpack image.

Recently that Renovate has been continually encountering the following error and has been unable to update dependencies that depend on Node.js v18.14.0. The same issue occurred when Node.js v18.13.0 was released.

```
DEBUG: Resolved stable matching version (repository=dummy/dummy-repo, baseBranch=develop, branch=renovate/dummy/dummy-2.x)
       "toolName": "npm",
       "constraint": "<9",
       "resolvedVersion": "8.19.3"
DEBUG: Executing command (repository=dummy/dummy-repo, baseBranch=develop, branch=renovate/dummy/dummy-2.x)
       "command": "install-tool node v18.14.0"
DEBUG: rawExec err (repository=dummy/dummy-repo, baseBranch=develop, branch=renovate/dummy/dummy-2.x)
       "err": {
         "name": "ExecError",
         "cmd": "/bin/sh -c install-tool node v18.14.0",
         "stderr": "No /opt/buildpack/tools/node/18.14.0/bin defined - aborting\n",
         "stdout": "tool node v18.14.0 is already installed\nlinking tool node v18.14.0\n",
         "options": {
           "cwd": "/tmp/renovate/github/dummy/dummy-repo/hoge/fuga/applications/system-admin",
           "encoding": "utf-8",
           "env": {
             "NPM_CONFIG_CACHE": "/tmp/renovate/cache/others/npm",
             "npm_config_store": "/tmp/renovate/cache/others/pnpm",
             "HOME": "/home/ubuntu",
             "PATH": "/home/ubuntu/.cargo/bin:/home/ubuntu/.local/bin:/go/bin:/opt/buildpack/tools/python/3.11.1/bin:/home/ubuntu/.npm-global/bin:/home/ubuntu/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
             "LC_ALL": "C.UTF-8",
             "LANG": "C.UTF-8",
[...]

PR #770, opened 2023-03-16
fix: create version folder after download

viceice:
most tools now create the version dir after download, so we've less dead installs

- #715

### DIFF
--- a/src/usr/local/buildpack/tools/v2/python.sh
+++ b/src/usr/local/buildpack/tools/v2/python.sh
@@ -46,21 +46,20 @@ function install_tool () {
     prepare_tool
   fi
 
-  versioned_tool_path=$(create_versioned_tool_path)
   arch=$(uname -p)
   url="https://github.com/containerbase/python-prebuild/releases/download"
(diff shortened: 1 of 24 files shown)


src/usr/local/buildpack/tools/v2/python.sh
```diff
@@ -46,22 +46,21 @@ function install_tool () {
...
   url="https://github.com/containerbase/python-prebuild/releases/download"
   version_codename=$(get_distro)
 
-  create_folder "${versioned_tool_path}/bin"
-
   file=$(get_from_url "${url}/${TOOL_VERSION}/python-${TOOL_VERSION}-${version_codename}-${arch}.tar.xz")
 
   if [[ -f ${file} ]]; then
-    bsdtar -C "${versioned_tool_path}" --strip 1 -xf "${file}"
+    bsdtar -C "${versioned_tool_path}" -xf "${file}"
```
viceice, line 56:
```suggestion
    bsdtar -C "$(find_tool_path)" -xf "${file}"
```

PR #854, opened 2023-03-30
chore: don't pin devcontainer to SHA

rarkins:


viceice (commented):
LGTM, but i should push a package rule to org preset first, so we don't get a new pin PR

rarkins:
Should this do it? https://github.com/containerbase/.github/pull/11

PR #870, opened 2023-04-03
feat(erlang): validate checksum

viceice:
we now create checksum files for our prebuild binaries


viceice:
yes, will remove them. i also need to check why some checksum files are missing for erlang

PR #973, opened 2023-05-12
feat(tools): add PDM to tools

secustor:
- Adds PDM for usage by the new PEP621 manager
- replaces h3py with h5py for tests as the former is no longer available

### DIFF
--- a/src/usr/local/buildpack/tools/v2/pdm.sh
+++ b/src/usr/local/buildpack/tools/v2/pdm.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+# shellcheck source=/dev/null
+. "$(get_buildpack_path)/utils/python.sh"
+
+function link_tool () {
+  post_install
+  [[ -n $SKIP_VERSION ]] || poetry --version
+}
--- a/src/usr/local/buildpack/tools/v2/pdm.sh
+++ b/src/usr/local/buildpack/tools/v2/pdm.sh
@@ -5,5 +5,5 @@
 
 function link_tool () {
   post_install
-  [[ -n $SKIP_VERSION ]] || poetry --version
+  [[ -n $SKIP_VERSION ]] || pdm --version
 }
--- a/test/Dockerfile.jammy
+++ b/test/Dockerfile.jammy
@@ -137,6 +137,8 @@ RUN install-tool pipenv 2023.4.29
 RUN install-tool hashin 0.17.0
 # renovate: datasource=pypi
 RUN install-tool poetry 1.4.2
+# renovate: datasource=pypi
+RUN install-tool pdm 2.6.1
 
 #--------------------------------------
 # Image: test-ruby
--- a/test/python/Dockerfile
+++ b/test/python/Dockerfile
@@ -200,10 +200,6 @@ FROM build as test-pdm
 
 USER 1000
 
-# make as latest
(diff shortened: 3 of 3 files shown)


src/usr/local/buildpack/tools/v2/pdm.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# shellcheck source=/dev/null
+. "$(get_buildpack_path)/utils/python.sh"
+
+function link_tool () {
+  post_install
+  [[ -n $SKIP_VERSION ]] || poetry --version
```
rarkins, line 8:
```suggestion
  [[ -n $SKIP_VERSION ]] || pdm --version
```

??

secustor: Copy and paste error

src/usr/local/buildpack/tools/v2/pdm.sh
rarkins, line 1: How/where is pdm actually installed here? I only see python

secustor: Maybe I'm understanding the functionality of base not correctly but my expectation has been that adding this file is enough to install it via pip. 🤔

rarkins: I've got to admit it looks that way in https://github.com/containerbase/base/blob/main/src/usr/local/buildpack/tools/v2/pipenv.sh for example. @viceice @Chumper is there some magic glue based on file name?

test/Dockerfile.jammy
```diff
@@ -137,6 +137,8 @@ RUN install-tool pipenv 2023.4.29
 RUN install-tool hashin 0.17.0
 # renovate: datasource=pypi
 RUN install-tool poetry 1.4.2
+# renovate: datasource=pypi
+RUN install-tool pdm 2.6.1
```
viceice, line 141: please also add it to python test dockerfile

test/python/Dockerfile
```diff
@@ -193,6 +193,20 @@ RUN set -ex; \
...
+#--------------------------------------
+# test pdm: pdm (multiple python)
+#--------------------------------------
+FROM build as test-pdm
+
+USER 1000
+
+# make as latest
+# renovate: datasource=github-releases packageName=containerbase/python-prebuild
+RUN install-tool python 3.11.3
```
viceice, line 205: isn't python already on build stage?

src/usr/local/buildpack/tools/v2/pdm.sh
viceice: yes, the python util defines the required functions and uses the passed env values to install the expected tool and version

test/python/Dockerfile
secustor: The tests for pipenv do also install python on each stage. https://github.com/containerbase/base/pull/973/files#diff-8722b87dfd4067380b8581bb8d4727b2bedaa8d620571984262fabc897af38b3R183

rarkins (approved):

test/python/Dockerfile
viceice: not all, only some to test older python. so remove to speed up tests

viceice (approved):

PR #998, opened 2023-05-25
feat: new tool installer

### LINKED ISSUE #771: Switch from bash scripts to higher language
We need to decide to which higher language we switch, because we've some features which are prettey complex to implement inside bash scripts.

So Idea's / preferences? 

---

- #101

viceice:
First draft. This simply forwards to the bash installer for now.

- closes #771


package.json
```diff
@@ -15,31 +15,43 @@
...
     "release:prepare": "node tools/prepare-release.js",
     "release:publish": "node tools/publish-release.js"
   },
-  "packageManager": "yarn@3.5.1"
+  "packageManager": "yarn@3.5.1",
+  "dependencies": {
+    "clipanion": "3.2.0",
+    "execa": "7.1.1",
+    "got": "12.6.0"
+  }
```
rarkins, line 56: Good to move these above `devDependencies`

viceice: it doesn't really matter, because we never publish this package. but they are production deps, because thex will be used in the build cli. 🤷‍♂️

rarkins:
Do you think you can make this so that it's a single command but with sub-commands? e.g. install, prepare, etc.?

viceice:
> Do you think you can make this so that it's a single command but with sub-commands? e.g. install, prepare, etc.?

I already did that with possible short link. So if you symlink the cli to `install-tool` you can call as now `install-tool node`.
otherwise you need to call `containerbase-cli install tool node`.

.yarnrc.yml
```diff
@@ -1,2 +1,14 @@
...
 defaultSemverRangePrefix: ''
+
+supportedArchitectures:
+  os:
+    - current
+    - linux
+    - win32
+  cpu:
+    - current
+    - x64
```
JamieMagee, line 11: Should you add `arm64` here?

tools/esbuild.js
```diff
@@ -0,0 +1,50 @@
+import { build } from 'esbuild';
+import { exec } from 'pkg';
+import esbuildPLuginPino from 'esbuild-plugin-pino';
```
JamieMagee, line 3: nit:
```suggestion
import esbuildPluginPino from 'esbuild-plugin-pino';
```

JamieMagee:
It looks like this is still CommonJS. What's preventing ESM?

tools/esbuild.js
JamieMagee, line 1: nit: this can be TypeScript as well.

src/cli/command/install-tools/legacy.ts
JamieMagee, line 1: Just to clarify: the idea is to use the CLI for everything straightaway. It'll pass through to existing shell scripts for tools that haven't been migrated, and scripts will be converted to TypeScript commands, before being deleted?

tools/esbuild.js
```diff
@@ -0,0 +1,50 @@
...
+import fs from 'node:fs/promises';
+
+await build({
+  entryPoints: { 'containerbase-cli': './src/cli/index.ts' },
+  bundle: true,
+  platform: 'node',
+  target: 'node18',
+  minify: false,
+  tsconfig: 'src/cli/tsconfig.json',
+  // format: "esm", // not supoorted
```
JamieMagee, line 13: What's the issue?

package.json
```diff
@@ -15,31 +15,49 @@
...
   "devDependencies": {
     "@semantic-release/exec": "6.0.3",
+    "@tsconfig/node18": "2.0.1",
     "@types/node": "18.16.13",
+    "@types/semver": "7.5.0",
     "@types/shelljs": "0.8.12",
     "bats": "1.9.0",
     "bats-assert": "2.0.0",
     "bats-support": "0.3.0",
     "commander": "10.0.1",
```
JamieMagee, line 25: Should we open an issue to migrate from `commander` to `clipanion` for existing scripts? (I think it's just [`tools/utils.js`][1])

[1]: https://github.com/containerbase/base/blob/main/tools/utils.js

.yarnrc.yml
viceice: yes, it seems so.

src/cli/command/install-tools/legacy.ts
viceice: yes, this PR should only bring some of the tools. more tools will be converted later

tools/esbuild.js
viceice: after compiling with pkg node will complain about dynamic import only allowed on esm. 

- https://github.com/vercel/pkg/issues/1291

package.json
viceice: sure

tools/esbuild.js
viceice: sure, but all other scripts are plain esm too :wink:

package.json
viceice: will do later after this PR

tools/esbuild.js
JamieMagee: True, but you can write ESM and have it converted to CJS by esbuild. I followed [this comment][1] to get it working in [my prototype][2]

[1]: https://github.com/vercel/pkg/issues/1291#issuecomment-1518694260
[2]: https://github.com/JamieMagee/containerbase-cli

JamieMagee:
Is this ready for review?

viceice:
> Is this ready for review?

no, cache is still missing. will try to do tomorrow

tools/esbuild.js
viceice: I've written as esm and esbuild is compiling to cjs by default because of `node18` target

PR #1015, opened 2023-05-30
feat!: move files from `buildpack` to `containerbase`

### LINKED ISSUE #826: move files from `<path>/buildpack` to `<path>/containerbase`
In a first step we can make a symlink to old place to not fully break exising users.
But will mark as `breaking` because of unknown side effects and better visibillity.

- #524

viceice:
Files are moved 
- `/usr/local/buildpack` to `/usr/local/containerbase`
- `/opt/buildpack` to `/opt/containerbase`

 `/usr/local/containerbase` will contain the tools and bash scripts.
`/opt/containerbase` will contain the generated files and installed tools.

- closes #826

```
BREAKING CHANGE: Containerbase files are moved, so check custom image samples for required changes
```


rarkins:
> Tools are still installed to `/opt/buildpack` because of overlapping folders 😕

Can you explain more? I was hoping to remove all `buildpack` remnants

Chumper (approved):

viceice:
will do another change.

source bash scripts and tools will be put to `/usr/local/containerbase` which is only writable by root and installed tools and other generated files to `/opt/containerbase`, which is writable by our custom user.

viceice:
@rarkins I've moved files again, see PR description. #1016 will be ladst `buildpack` piece which get's removed in the following major (probably v9)

PR #1210, opened 2023-07-18
feat: support url rewrite

### LINKED ISSUE #1067: Allow all external download URLs to be aliased
Some users need to build their own containerbase-derived images in their product environments which do not have direct internet access. Instead, they need to set up generic proxies on Artifactory for each distinct external host which is approved, such as github.com, nodejs.org, etc.

Here's some examples used by containerbase:

Host | Tool(s)
-- | --
https://nodejs.org | node
https://api.adoptium.net | java
https://downloads.lightbend.com | scala
https://github.com | sbt, python, php,
https://dot.net | dotnet
https://cache.ruby-lang.org/ | ruby

We can assume for now that such users can already redirect common registries like apt, npm, pip, etc - the primary concern is on arbitrary URLs like the above.

The best way to test this would be to have a build environment with external URLs blocked by default and then try to build a "full" image with all tools and find the missing URLs one by one.

The next thing we'd want is a simple way of defining alternative URLs using env.

One possibility would be to have a syntax where the full from/to is specified in separate, related variables. e.g.

`URL_REPLACE_0_FROM=https://node.org`
`URL_REPLACE_0_TO=https://artifactory.company.com/something/nested`

Such a syntax is verbose but then at least should cover every case.

A higher level approach would be like `JAVA_BASE_URL=https://artifactory.company.com/something/java` but has the downsides:
- One tool might try multiple hosts, and
- Multiple tools might use the same host (e.g. github.com)

---

**Supported Tools**
- [x] `docker` #1189 
- [x] `dart` #1197

viceice:
We now call back into our new cli to download the file, so the new url replacement can be applied.
Will add some more documentation too later.

- closes #1067
- https://github.com/vercel/pkg/issues/1861


src/usr/local/containerbase/utils/cache.sh
```diff
@@ -90,8 +90,9 @@ function download_file () {
   local temp_folder=${CONTAINERBASE_CACHE_DIR:-${TEMP_DIR}}
   while [ "${retry}" -gt 0 ]; do
     retry=$((retry-1))
-    if ! curl --retry 3 --create-dirs -sSfLo "${temp_folder}/${name}" "${url}" ; then
+    if ! containerbase-cli df "${url}" "${temp_folder}/${name}" >&2; then
       echo "Download failed: ${url}" >&2
+      env >&2
```
viceice, line 95:
```suggestion
```

ups

Chumper (approved):

Chumper:
We can phase out the bats tests whenever we replace functionality with typescript

viceice:
> We can phase out the bats tests whenever we replace functionality with typescript

fixed the tests. I'm not sure if i need the file delete 🤔

src/cli/services/env.service.ts
```diff
@@ -91,7 +91,7 @@ export class EnvService {
 
     return (this.replacements = replacements);
   }
-
+
```
viceice, line 94:
```suggestion
```

README.md
```diff
@@ -98,10 +98,9 @@ You can replace the default urls used to download the tools.
 The number of `URL_REPLACE_*_FROM` and `URL_REPLACE_*_TO` environment variables must match.
 Theses variables are case sensitive.
 The numbers will be processed in numerical order and can have gaps.
-This is currently only supported by these tool installers:
-
-- `docker`
-- `dart`
+Almost all tools are noew supported.
```
viceice, line 101:
```suggestion
Almost all tools are now supported.
```

README.md
```diff
@@ -98,10 +98,9 @@ You can replace the default urls used to download the tools.
...
+Only tooles which are installed by `gem`, `npm` or `pip` are not supported.
```
viceice, line 102:
```suggestion
Only tools which are installed by `gem`, `npm` or `pip` are not supported.
```

rarkins (changes requested):
Can we get docs for the URL rewriting if this closes the issue?

viceice:
> Can we get docs for the URL rewriting if this closes the issue?

there are already some docs and all urls are printed on debug log level

viceice:
> P My

???

rarkins:
Not sure where that came from. Can you point me to the docs?

viceice:
> Not sure where that came from. Can you point me to the docs?

https://github.com/containerbase/base/tree/feat/url-replace#url-replacement

will create a mother extended version in later PR's.

will try to extract the base urls to variables, so I can use a test to validate docs to have a sample for each of them.

rarkins (approved):
Don't want to block this but we need conclusive docs for rewriting all URLs in a "full" image

viceice:
yes, will update docs later. currently it's best to try and error to see failed urls in logs. then it should be easy to rewrite them.

PR #1291, opened 2023-08-09
feat(node): enable use-openssl-ca

viceice:
Using `NODE_EXTRA_CA_CERTS` is no longer required.

Using `NODE_OPTIONS=--use-openssl-ca` is known to be broken on Node `>=18.0.0 <18.5.0`[^1]

[^1]: https://nodejs.org/en/blog/vulnerability/july-2022-security-releases#openssl---aes-ocb-fails-to-encrypt-some-bytes-mediumcve-2022-2097

### DIFF
--- a/docs/custom-root-ca.md
+++ b/docs/custom-root-ca.md
@@ -13,9 +13,6 @@ FROM containerbase/base
 
 COPY my-root-ca.crt /usr/local/share/ca-certificates/my-root-ca.crt
 RUN update-ca-certificates
-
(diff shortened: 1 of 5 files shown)


rarkins:
Does this change how end users configure certs?

viceice:
> Does this change how end users configure certs?

Not really, they can skip `NODE_EXTRA_CA_CERTS`, but it will still work with it. So no breaking here.

PR #1663, opened 2023-10-28
test(node): update tests for v20 LTS plus v16 EOL

rarkins:


test/node/Dockerfile
```diff
@@ -190,7 +190,7 @@ RUN install-tool node 10.24.0
 # intentionally fails
 RUN install-tool npm 9.5.1 || true
 RUN install-tool yarn-slim 1.22.19
-RUN install-tool node v18.16.1
+RUN install-tool node v20.9.0
```
viceice, line 193: it seems yarn v1 has issues with node v20

viceice:
there's another re2 test case which needs an update. will do it as soon as possible

viceice (approved):

PR #1737, opened 2023-11-15
feat(tools): add bazelisk

aschleifer:
closes https://github.com/containerbase/base/issues/20

This adds `bazelisk` as possible tool to the containerbase. That way it can be used in renovate.


viceice (commented):
can you please don't like flux, so I don't need to convert it after merge?

https://github.com/containerbase/base/blob/main/src/cli/tools/flux/index.ts

aschleifer:
> can you please don't like flux, so I don't need to convert it after merge?
> 
> https://github.com/containerbase/base/blob/main/src/cli/tools/flux/index.ts

Hello, do I understand you correctly that you want me to do it in typescript? If so and flux is the example then I can do it next week for sure.

viceice:
yes, thanks 🙏

aschleifer:
Hello @viceice ,

I rewrote the tool in typescript.

As an improvement for the future it might help to have some guide explaining how to to do it. I found the other file changes (not in the `src/cli/tools/bazelisk` folder) out by grepping for "flux" and reading the code. 

Also I didn't find any documentation stating what the "prepare" stage is suppose to be used for.

viceice (commented):
please add a test install 

https://github.com/containerbase/base/blob/cddf64e46c78e129fa5a9942064a9f5ebdf5fa75/test/Dockerfile.jammy#L33

https://github.com/containerbase/base/blob/cddf64e46c78e129fa5a9942064a9f5ebdf5fa75/test/latest/Dockerfile.arm64#L25

viceice:
please don't force push

src/cli/tools/bazelisk/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    });
+  }
+
+  override async link(version: string): Promise<void> {
+    const src = join(this.pathSvc.versionedToolPath(this.name, version), 'bin');
+
+    await this.shellwrapper({
+      srcDir: src,
+    });
+    await fs.symlink('bazelisk', join(this.pathSvc.binDir, 'bazel'));
```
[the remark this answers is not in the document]
aschleifer: why does this look wrong? `bazel` should be using the `bazelisk` executable.

aschleifer:
> please add a test install
> 
> https://github.com/containerbase/base/blob/cddf64e46c78e129fa5a9942064a9f5ebdf5fa75/test/Dockerfile.jammy#L33
> 
> https://github.com/containerbase/base/blob/cddf64e46c78e129fa5a9942064a9f5ebdf5fa75/test/latest/Dockerfile.arm64#L25

I added some tests. Let me know if something is missing.

src/cli/tools/bazelisk/index.ts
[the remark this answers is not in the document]
viceice: sure, but the first argument of node fs symlink needs to be the full path or relative to cwd (which is unknown and can be arbitrary at runtime)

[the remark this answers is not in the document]
aschleifer: > sure, but the first argument of node fs symlink needs to be the full path or relative to cwd (which is unknown and can be arbitrary at runtime)

Good point. I changed the symlink to be absolute now.

test/latest/Dockerfile.arm64
```diff
@@ -16,6 +16,15 @@ COPY src/ /
...
 RUN install-containerbase
 
+#--------------------------------------
+# Image: bazelisk
+#--------------------------------------
+FROM base as test-bazelisk
+
+
+# renovate: datasource=github-tags packageName=bazelbuild/bazelisk
+RUN install-tool bazelisk v1.18.0
```
viceice, line 26:
```suggestion
RUN install-tool bazelisk v1.18.0
RUN bazel --version
```

test symlink 🙃

src/cli/tools/bazelisk/index.ts
```diff
@@ -0,0 +1,65 @@
...
+    @inject(EnvService) envSvc: EnvService,
+    @inject(PathService) pathSvc: PathService,
+    @inject(HttpService) private http: HttpService,
+  ) {
+    super(pathSvc, envSvc);
+  }
+
+  override async install(version: string): Promise<void> {
+    const baseurl = `https://github.com/bazelbuild/bazelisk/releases/download/v${version}/`;
+    const filename = `bazelisk-linux-${this.arch}`;
```
viceice, line 26:
```suggestion
    const filename = `bazelisk-linux-${this.envSvc.arch}`;
```

src/cli/tools/bazelisk/index.ts
```diff
@@ -0,0 +1,65 @@
...
+import { EnvService, HttpService, PathService } from '../../services';
+
+@injectable()
+export class InstallBazeliskService extends InstallToolBaseService {
+  readonly name = 'bazelisk';
+
+  private get arch(): string {
+    return this.envSvc.arch;
+  }
+
```
viceice, line 15:
```suggestion
```
We don't need this simple property

test/latest/Dockerfile.arm64
```diff
@@ -16,6 +16,16 @@ COPY src/ /
...
 
+#--------------------------------------
+# Image: bazelisk
+#--------------------------------------
+FROM base as test-bazelisk
+
+
+# renovate: datasource=github-releases packageName=bazelbuild/bazelisk
+RUN install-tool bazelisk v1.18.0
+RUN bazel --version
```
viceice, line 27:
```suggestion
```

viceice (approved):

PR #1991, opened 2024-01-15
feat!: update ubuntu to 22.04 Jammy Jellyfish

viceice:
Update base image to jammy

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -149,7 +149,7 @@ jobs:
     strategy:
       matrix:
         distro:
-          - jammy
+          - focal
 
     env:
       TAG: ${{ matrix.distro }}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 #--------------------------------------
 # Image: base
 #--------------------------------------
-FROM ubuntu:20.04@sha256:f2034e7195f61334e6caff6ecf2e965f92d11e888309065da85ff50c617732b8
+FROM ubuntu:jammy@sha256:6042500cf4b44023ea1894effe7890666b0c5c7871ed83a97c36c76ae560bb9b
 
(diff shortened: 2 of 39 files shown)


rarkins (changes requested):
Pending internal review/approval at Mend

.github/workflows/build.yml
```diff
@@ -149,7 +149,7 @@ jobs:
     strategy:
       matrix:
         distro:
-          - jammy
+          - focal
```
HonkingGoose, line 152: Focal Fossa is a _downgrade_ from Jammy Jellyfish. Is that what you really want?

Here's the recent list of Ubuntu branding names for reference:

- 24.04 Noble Numbat
- 22.04 Jammy Jellyfish
- 20.04 Focal Fossa
- 18.04 Bionic Beaver

viceice: yes, because we flipped the default, but still support the old version 😉

HonkingGoose: Ah okay! :smile: You can resolve this review comment then. :wink:

Dockerfile
```diff
@@ -1,7 +1,7 @@
 #--------------------------------------
 # Image: base
 #--------------------------------------
-FROM ghcr.io/containerbase/ubuntu:20.04@sha256:0b897358ff6624825fb50d20ffb605ab0eaea77ced0adb8c6a4b756513dec6fc
+FROM ghcr.io/containerbase/ubuntu:24.04@sha256:2e863c44b718727c860746568e1d54afd13b2fa71b160f5cd9058fc436217b30
```
viceice, line 4:
```suggestion
FROM ghcr.io/containerbase/ubuntu:24.04@sha256:b359f1067efa76f37863778f7b6d0e8d911e3ee8efa807ad01fbf5dc1ef9006b
```

PR #2145, opened 2024-02-06
feat(python): Add keyrings.envvars package to pip-tools

mbudnek:
This goes along with https://github.com/renovatebot/renovate/pull/26853.  @viceice Raised the concern credentials passed as part of index URLs can be written into pip-compile's output file, which could easily leak plain-text credentials.

To avoid that, I want to use the keyrings.envvars package to pass credentials to pip-compile via the keyring module.  Credentials that pip/pip-compile looks up via the keyring module don't ever get written to pip-compile's output file.


src/usr/local/containerbase/tools/v2/poetry.sh
```diff
@@ -3,6 +3,10 @@
 # shellcheck source=/dev/null
 . "$(get_containerbase_path)/utils/python.sh"
 
+function install_tool () {
+  install_python_tool
+}
+
```
viceice, line 9: move this to utils/python.sh as default, so it only needs to be over written where needed.

src/usr/local/containerbase/tools/v2/pip-tools.sh
```diff
@@ -3,6 +3,30 @@
...
+  export PYTHON_MINOR=${BASH_REMATCH[3]}
+}
+
+function install_tool () {
+  local python_version
+
+  python_version=$(get_tool_version python)
+  split_python_version "${python_version}"
+
+  if [[ ${PYTHON_MAJOR} == 3 && ${PYTHON_MINOR} -ge 9 ]]; then
```
viceice, line 22: this would break if python v4 will be released

viceice (changes requested):
please extract the refactoring to a separate PR to be merged first.

mbudnek:
@viceice I opened #2149 for the refactor

src/usr/local/containerbase/tools/v2/pip-tools.sh
```diff
@@ -3,6 +3,30 @@
...
+}
+
+function install_tool () {
+  local python_version
+
+  python_version=$(get_tool_version python)
+  split_python_version "${python_version}"
+
+  if [[ (${PYTHON_MAJOR} == 3 && ${PYTHON_MINOR} -ge 9) || ${PYTHON_MAJOR} -gt 3 ]]; then
+      install_python_tool 'keyrings.envvars >= 1.1.0'
```
viceice, line 23: Why do we need this version contraint: `keyrings.envvars >= 1.1.0`

viceice (commented):
We need to check on renovate:

If there is a python constaint extracted validate it allows python 3.9+ otherwise war that we can't use `keyrings.envvars`

src/usr/local/containerbase/tools/v2/pip-tools.sh
viceice: so remove the constraints and let it install latest version?

mbudnek: The constraint is just to make sure it doesn't install an _older_ version that doesn't have the functionality Renovate will need.  I can remove it if you would prefer, but I think having it there serves as an indication that something changed in that version that we need.  I can leave a comment to that effect if the objection is just the unexplained magic number.

viceice: ok, let's wait with this PR until your feature is released and we know the minimal version to use

mbudnek: I added a comment explaining why the version constraint is needed

mbudnek: Also, the new version of keyrings.envvars with my changes is now available, so this should be ready for review/merge.

test/python/Dockerfile.arm64
```diff
@@ -43,6 +43,14 @@ FROM test-python as test-hashin
...
 RUN install-tool hashin 0.17.0
 
+#--------------------------------------
+# Image: pip-tools
+#--------------------------------------
+FROM test-python as test-pip-tools
+
+# renovate: datasource=pypi
+RUN install-tool pip-tools 7.4.0
+
```
viceice, line 53: remove, arm64 emulation is too slow, so only do really required things

viceice (approved):

PR #2149, opened 2024-02-06
refactor(python): Accept optional extra packages to install

mbudnek:
I'm planning to add the keyrings.envvars package to be installed with the pip-tools tool, so instead of copy/pasting the common python install_tool logic a second time this refactors it into a common function that will install any packages given as arguments along with the tool being installed.


src/usr/local/containerbase/utils/python.sh
```diff
@@ -30,7 +30,8 @@ function check_tool_installed() {
   test -n "$(find_pip_versioned_path)"
 }
 
-function install_tool() {
+# shellcheck disable=2120
```
viceice, line 33:
```suggestion
# shellcheck disable=SC2120
```
I think we use the prefix on other places

viceice (commented):
otherwise LGTM

viceice (approved):
next time use the re-request review button next to the pr reviewers 😉

viceice:
please don't merge main, branch doesn't have to be up to date. we use merge queue which will do a squash merge and test it before merging to target.

viceice (approved):

mbudnek:
@viceice I merged main just to get the tests to re-run, since some sort of docker daemon issue caused them to fail.

viceice:
ok, ping me next time. I can simply rerun only failed jobs, which is faster than a full run

PR #2684, opened 2024-05-23
feat: add rye

vhdirk:
This MR adds support for [rye](https://github.com/astral-sh/rye), as per https://github.com/renovatebot/renovate/issues/25273#issuecomment-1768042170 

Everything should be here, but for some reason I can't get it to pick up `TOOL_VERSION`? Not sure what's going on there


src/usr/local/containerbase/tools/v2/rye.sh
```diff
@@ -0,0 +1,85 @@
...
+
+  file=$(get_from_url "${file_url}" "${file_name}" "${expected_checksum}" "sha256sum")
+
+  versioned_tool_path=$(create_versioned_tool_path)
+  create_folder "${versioned_tool_path}/bin"
+
+  gunzip -c "${file}" > "${versioned_tool_path}/bin/rye"
+  chmod +x "${versioned_tool_path}/bin/rye"
+
+  "${versioned_tool_path}/bin/rye" self install --yes
```
viceice, line 76: what is this doing?

test/Dockerfile.jammy
```diff
@@ -151,7 +151,9 @@ RUN install-tool pipenv 2023.12.1
 RUN install-tool pdm 2.15.3
 # renovate: datasource=pypi
 RUN install-tool poetry 1.8.3
-
+RUN prepare-tool rye
```
viceice, line 154:
```suggestion
```
already done above

test/Dockerfile.noble
```diff
@@ -151,7 +151,9 @@ RUN install-tool pipenv 2023.12.1
 RUN install-tool pdm 2.15.3
 # renovate: datasource=pypi
 RUN install-tool poetry 1.8.3
-
+RUN prepare-tool rye
```
viceice, line 154:
```suggestion
```

test/python/Dockerfile
```diff
@@ -72,6 +72,17 @@ RUN install-tool poetry 1.8.3
 
 USER 1000
 
+#--------------------------------------
+# build: rye
+#--------------------------------------
+FROM build as rye
+
+RUN prepare-tool rye
```
viceice, line 80: add to line 48

test/python/Dockerfile
```diff
@@ -247,6 +258,21 @@ USER 1000
 # renovate: datasource=pypi
 RUN install-tool pdm 2.15.3
 
+#--------------------------------------
+# test: rye: rye (multiple python)
+#--------------------------------------
+FROM build as test-rye
+
+RUN prepare-tool rye
```
viceice, line 266: see other comment

test/python/Dockerfile
```diff
@@ -247,6 +258,21 @@ USER 1000
 # renovate: datasource=pypi
 RUN install-tool pdm 2.15.3
 
+#--------------------------------------
+# test: rye: rye (multiple python)
```
viceice, line 262: ? i don't see multiple python. should be rootless indstead?

test/python/Dockerfile.arm64
```diff
@@ -59,6 +59,17 @@ FROM test-python as test-poetry
 # renovate: datasource=pypi
 RUN install-tool poetry 1.8.3
 
+#--------------------------------------
+# build: rye
+#--------------------------------------
+FROM test-python as test-rye
+
+RUN prepare-tool rye
```
viceice, line 67:
```suggestion
```

will be implicit called

viceice (changes requested):
Does `rye` require `python`?

https://github.com/containerbase/base/blob/290455ad82452f22f14051bd2557783a15c66603/src/usr/local/containerbase/utils/python.sh#L3-L6

you should a test which calls rye like renovate will do later. See pipenv and poetry tests

src/usr/local/containerbase/tools/v2/rye.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+
+function prepare_tool() {
+  local version_codename
+  local tool_path
+
+  version_codename="$(get_distro)"
+  case "${version_codename}" in
+    "focal" | "jammy" | "noble" )
+      apt_install gzip
```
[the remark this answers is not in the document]
vhdirk: The github release is provided as *.gz rather than *.tar.gz. I am not sure if gzip was already installed

src/usr/local/containerbase/tools/v2/rye.sh
vhdirk: rye needs to bootstrap itself. It installs itself to (by default) to `~/.rye` where it stores a copy of itself, after which `<tool_path>/bin/rye` may be removed.
In `~/.rye`, it also stores different versions of python (if needed) and additional tools like `uv` and `ruff` (if needed), as well as a pip registry cache for packages that have been used before.

test/python/Dockerfile
vhdirk: Ah, yes, indeed. will fix asap

test/python/Dockerfile
vhdirk: What do you mean?

src/usr/local/containerbase/tools/v2/rye.sh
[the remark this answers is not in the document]
viceice: comment out and try without

src/usr/local/containerbase/tools/v2/rye.sh
viceice: can that folder be configured? we like to support read-only root filesystem, so this is working against it.

can't we simply run it from the installed folder?

test/python/Dockerfile
viceice: use a single prepare tool call with multiple tool names

src/usr/local/containerbase/tools/v2/rye.sh
vhdirk: Yes, it is configurable, so it is possible to install it into the `versioned_tool_path`. It does expect read/write permissions in that dir

test/python/Dockerfile
vhdirk: Also, Rye can be configured to fetch a specific version of the python interpreter based on the pyproject file it encounters. This is the default setting. As there's actually no dependency on python at all. I just put it under python since that's what its intended purpose is.

src/usr/local/containerbase/tools/v2/rye.sh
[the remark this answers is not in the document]
vhdirk: Works without :)

src/usr/local/containerbase/tools/v2/rye.sh
```diff
@@ -0,0 +1,85 @@
...
+    if [[ $(is_root) -ne 0 ]]; then
+      echo "${TOOL_NAME} not prepared"
+      exit 1
+    fi
+    prepare_tool
+    tool_path=$(find_tool_path)
+  fi
+
+  if [[ $platform == "Darwin" ]]; then
+    platform="macos"
```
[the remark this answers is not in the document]
vhdirk: Removed

vhdirk:
@viceice  I've moved Rye to it's own container since it doesn't need python at all. It manages all of that by itself.

test/rye/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+SHELL [ "/bin/sh", "-c" ]
+RUN rye --version
+RUN rye run python --version
+
+#--------------------------------------
+# test a2: non-root
+#--------------------------------------
+FROM build as test-a2
+
+USER 1000
```
viceice, line 54: you should install rye as user here to have a better test

viceice:
@vhdirk Please resolve fixed conversations 🤗

test/rye/Dockerfile
vhdirk: Done :)

src/usr/local/containerbase/tools/v2/rye.sh
vhdirk: Rye will now install itself to the versioned tool path

docs/custom-registries.md
```diff
@@ -526,6 +526,19 @@ https://static.rust-lang.org/dist/rust-1.75.0-aarch64-unknown-linux-gnu.tar.gz
...
+
+Rye releases are downloaded from:
+
+- `https://github.com/astral-sh/rye/releases`
+
+Samples:
+
+```txt
+https://github.com/astral-sh/rye/releases/download/0.34.0/rye-x86_64-linux.gz.sha256
+https://github.com/astral-sh/rye/releases/download/0.34.0/rye-x86_64-linux.gz
```
viceice, line 539: please add an arm64 sample

src/usr/local/containerbase/tools/v2/rye.sh
```diff
@@ -0,0 +1,65 @@
...
+
+  if [[ $platform == "Linux" ]]; then
+    platform="linux"
+  fi
+
+  if [[ $arch == armv8* ]] || [[ $arch == arm64* ]] || [[ $arch == aarch64* ]]; then
+    arch="aarch64"
+  elif [[ $arch == i686* ]]; then
+    arch="x86"
+  fi
```
viceice, line 25: remove, we never support non x64 systems.

src/usr/local/containerbase/tools/v2/rye.sh
```diff
@@ -0,0 +1,65 @@
...
+  name=$TOOL_NAME
+  version=$TOOL_VERSION
+  platform=$(uname -s)
+  repo="astral-sh/$name"
+
+  if [[ $platform == "Linux" ]]; then
+    platform="linux"
+  fi
+
+  if [[ $arch == armv8* ]] || [[ $arch == arm64* ]] || [[ $arch == aarch64* ]]; then
```
viceice, line 21: please check other installers and simplify, architecture can only have 2 values.

src/usr/local/containerbase/tools/v2/rye.sh
```diff
@@ -0,0 +1,65 @@
...
+
+  arch=${ARCHITECTURE}
+  name=$TOOL_NAME
+  version=$TOOL_VERSION
+  platform=$(uname -s)
+  repo="astral-sh/$name"
+
+  if [[ $platform == "Linux" ]]; then
+    platform="linux"
+  fi
```
viceice, line 19: remove, platform is always linux

src/usr/local/containerbase/tools/v2/rye.sh
```diff
@@ -0,0 +1,65 @@
...
+  checksum_file=$(get_from_url "${file_url}.sha256")
+  # get checksum from file
+  expected_checksum=$(cut -d' ' -f1  "${checksum_file}")
+
+  file=$(get_from_url "${file_url}" "${file_name}" "${expected_checksum}" "sha256sum")
+
+  versioned_tool_path=$(create_versioned_tool_path)
+  create_folder "${versioned_tool_path}/bin"
+
+  gunzip -c "${file}" > "${TEMP_DIR}/rye"
```
viceice, line 41: use bsdtar

src/usr/local/containerbase/tools/v2/rye.sh
```diff
@@ -0,0 +1,65 @@
...
+  chmod +x "${TEMP_DIR}/rye"
+
+  export RYE_HOME="${versioned_tool_path}"
+  "${TEMP_DIR}/rye" self install --yes
```
viceice, line 45: is this downloading additional files? what the effects?

viceice (commented):
can rye reuse existing installed python versions? so we can better control and cache it?

is rye downloading pre compiled python or is it compiling on the fly / first time usage?

viceice:
@rarkins rye is installing python on the fly when invoking. it's downloading a pre build python version from GitHub. this can cause issues on air gapped renovate installations.
the source doesn't seem to be configurable.

but we can register custom python versions, so rye can use it instead of downloading one on the fly.

should we disable auto install by default for containerbase?

I also think if we should use those pre build python versions instead of building ourselves.
just republish the binaries to our prebuild repo for easier installs.

https://gregoryszorc.com/docs/python-build-standalone/main/

vhdirk:
If the source needs to be configurable, I could open an issue/PR on rye to enable it to do that? 
Rye should be configurable to use the system python, though it does seem to download _a_ python version anyway? (That might be something to do with my system, though)

rarkins:
Ideally could be configured to never download python itself

vhdirk:
It seems to be possible to configure toolchains (https://rye.astral.sh/guide/toolchains/#registering-toolchains). However, any project that pinned a toolchain with a `.python-version` is able to override this (https://rye.astral.sh/guide/toolchains/#pinning-toolchains).
We could also just prefetch every downloadable toolchain upfront: https://rye.astral.sh/guide/toolchains/#fetching-toolchains ?

viceice:
> It seems to be possible to configure toolchains (https://rye.astral.sh/guide/toolchains/#registering-toolchains). However, any project that pinned a toolchain with a `.python-version` is able to override this (https://rye.astral.sh/guide/toolchains/#pinning-toolchains).

that's why I think we should download the python version ourselves and register it. disabling auto downloads.


> We could also just prefetch every downloadable toolchain upfront: https://rye.astral.sh/guide/toolchains/#fetching-toolchains ? 

too much and too slow, because most of the time we only need one chain

rarkins:
Ideally rye is configurable to never download python and use the system python instead

src/usr/local/containerbase/tools/v2/rye.sh
vhdirk: Can't use bsdtar: the file is not a `tar.gz`, just a plain `gz`

vhdirk:
> Ideally rye is configurable to never download python and use the system python instead

If the project under consideration (for renovate, that is) pinned a python version using `.python-version`, rye will _always_ download that python version if it doesn't match the system python version. Even when only locking and not actually installing the dependencies.

viceice:
> > Ideally rye is configurable to never download python and use the system python instead
> 
> If the project under consideration (for renovate, that is) pinned a python version using `.python-version`, rye will _always_ download that python version if it doesn't match the system python version. Even when only locking and not actually installing the dependencies.
> 
> 

but it can be disabled and we can already ensure a proper python version.

docs/custom-registries.md
bvdbasch: @vhdirk I already looked them up for you to save you a few minutes ;)

`https://github.com/astral-sh/rye/releases/download/0.34.0/rye-aarch64-linux.gz`
`https://github.com/astral-sh/rye/releases/download/0.34.0/rye-x86_64-linux.gz.sha256`

src/usr/local/containerbase/tools/v2/rye.sh
bvdbasch: @viceice From what I can gather this is at the final step in the "official installer". This "installer" is a script which you are supposed to pipe directly into bash/zsh/etc. I don't think that's a great way to install things, so I'm happy @vhdirk implemented it without that mechanism. You can find the curl into shell command on the official Rye site here: https://rye.astral.sh

I have also copied the current version of the script below for archival purposes should anyone stumble upon this PR in the future :) **Note the `self install` command on line 80**

```bash
    1 #!/usr/bin/env bash
    2 set -euo pipefail
    3 
    4 # Wrap everything in a function so that a truncated script
    5 # does not have the chance to cause issues.
    6 __wrap__() {
    7 
    8 # allow overriding the version
    9 VERSION=${RYE_VERSION:-latest}
   10 # allow overriding the install option
   11 INSTALL_OPTION=${RYE_INSTALL_OPTION:-""}
   12 
   13 REPO=astral-sh/rye
   14 PLATFORM=`uname -s`
   15 ARCH=`uname -m`
   16 
   17 if [[ $PLATFORM == "Darwin" ]]; then
   18   PLATFORM="macos"
   19 elif [[ $PLATFORM == "Linux" ]]; then
   20   PLATFORM="linux"
   21 fi
   22 
   23 if [[ $ARCH == armv8* ]] || [[ $ARCH == arm64* ]] || [[ $ARCH == aarch64* ]]; then
   24   ARCH="aarch64"
   25 elif [[ $ARCH == i686* ]]; then
   26   ARCH="x86"
   27 fi
   28 
   29 BINARY="rye-${ARCH}-${PLATFORM}"
   30 
   31 # Oddly enough GitHub has different URLs for latest vs specific version
   32 if [[ $VERSION == "latest" ]]; then
   33   DOWNLOAD_URL=https://github.com/${REPO}/releases/latest/download/${BINARY}.gz
   34 else
   35   DOWNLOAD_URL=https://github.com/${REPO}/releases/download/${VERSION}/${BINARY}.gz
   36 fi
   37 
   38 echo "This script will automatically download and install rye (${VERSION}) for you."
   39 if [ "x$(id -u)" == "x0" ]; then
   40   echo "warning: this script is running as root.  This is dangerous and unnecessary!"
   41 fi
   42 
   43 if ! hash curl 2> /dev/null; then
   44   echo "error: you do not have 'curl' installed which is required for this script."
   45   exit 1
   46 fi
   47 
   48 if ! hash gunzip 2> /dev/null; then
   49   echo "error: you do not have 'gunzip' installed which is required for this script."
   50   exit 1
   51 fi
   52 
   53 TEMP_FILE=`mktemp "${TMPDIR:-/tmp}/.ryeinstall.XXXXXXXX"`
   54 TEMP_FILE_GZ="${TEMP_FILE}.gz"
   55 
   56 cleanup() {
   57   rm -f "$TEMP_FILE"
   58   rm -f "$TEMP_FILE_GZ"
   59 }
   60 
   61 trap cleanup EXIT
   62 HTTP_CODE=$(curl -SL --progress-bar "$DOWNLOAD_URL" --output "$TEMP_FILE_GZ" --write-out "%{http_code}")
   63 if [[ ${HTTP_CODE} -lt 200 || ${HTTP_CODE} -gt 299 ]]; then
   64   echo "error: platform ${PLATFORM} (${ARCH}) is unsupported."
   65   exit 1
   66 fi
   67 
   68 rm -f "$TEMP_FILE"
   69 gunzip "$TEMP_FILE_GZ" 
   70 chmod +x "$TEMP_FILE"
   71 
   72 # Detect when the file cannot be executed due to NOEXEC /tmp.  Taken from rustup
   73 # https://github.com/rust-lang/rustup/blob/87fa15d13e3778733d5d66058e5de4309c27317b/rustup-init.sh#L158-L159
   74 if [ ! -x "$TEMP_FILE" ]; then
   75   printf '%s\n' "Cannot execute $TEMP_FILE (likely because of mounting /tmp as noexec)." 1>&2
   76   printf '%s\n' "Please copy the file to a location where you can execute binaries and run it manually." 1>&2
   77   exit 1
   78 fi
   79 
   80 "$TEMP_FILE" self install $INSTALL_OPTION
   81 
   82 }; __wrap__
```

bvdbasch:
I am patiently awaiting rye support in Renovate. To help speed this along I've looked at the latest unresolved conversations and provided answers to the questions raised by @viceice. Hope it helps :)

Thanks @vhdirk for all the work you've put into this!

Chippiewill:
I've had a bit of a spelunk in the Rye codebase. It doesn't seem possible at this time to prevent Rye from installing a Python version that's not there (there's no configuration or option check in that codepath), you can only prevent it by registering it. That should be straight forward, but is a bit more fragile. Would that be reasonable?

Strictly speaking Rye shouldn't need _any_ Python installation to update the lockfiles. It actually does that all by itself. Rye currently only install Python so it can initialise the virtualenv, it initialises the virtualenv because the `rye lock` command is actually just a simple wrapper around `rye sync` that would ordinarily install the locked packages as well. It seems a better fix for this would be to make an improvement in Rye upstream so that locking _only_ doesn't create a virtual env and doesn't install Python at all.

viceice:
> I've had a bit of a spelunk in the Rye codebase. It doesn't seem possible at this time to prevent Rye from installing a Python version that's not there (there's no configuration or option check in that codepath), you can only prevent it by registering it. That should be straight forward, but is a bit more fragile. Would that be reasonable?

That was my idea, yes. 

> Strictly speaking Rye shouldn't need _any_ Python installation to update the lockfiles. It actually does that all by itself (or rather it uses uv to do it). Rye currently only install Python so it can initialise the virtualenv, it initialises the virtualenv because the `rye lock` command is actually just a simple wrapper around `rye sync` that would ordinarily install the locked packages as well. It seems a better fix for this would be to make an improvement in Rye upstream so that locking _only_ doesn't create a virtual env and doesn't install Python at all.

That's even a much better option / solution.

docs/custom-registries.md
viceice: please add them to the docs

src/usr/local/containerbase/tools/v2/rye.sh
```diff
@@ -0,0 +1,58 @@
...
+  bsdtar -x -f "${file}" -C "${TEMP_DIR}/rye"
+  chmod +x "${TEMP_DIR}/rye"
+
+  export RYE_HOME="${versioned_tool_path}"
+  "${TEMP_DIR}/rye" self install --yes --no-modify-path
+
+  # Fetch every conceivable toolchain
+  for toolchain in $(rye toolchain list --include-downloadable |grep downloadable | cut -d ' ' -f 1); do
+    rye toolchain fetch "${toolchain}"
+  done
```
viceice, line 38: no, remove

src/usr/local/containerbase/tools/v2/rye.sh
```diff
@@ -0,0 +1,58 @@
...
+  checksum_file=$(get_from_url "${file_url}.sha256")
+  # get checksum from file
+  expected_checksum=$(cut -d' ' -f1  "${checksum_file}")
+
+  file=$(get_from_url "${file_url}" "${file_name}" "${expected_checksum}" "sha256sum")
+
+  versioned_tool_path=$(create_versioned_tool_path)
+  create_folder "${versioned_tool_path}/bin"
+
+  bsdtar -x -f "${file}" -C "${TEMP_DIR}/rye"
```
viceice, line 29: ci is failing because of corrupt / unreadable archive

vhdirk: That's because bsdtar doesn't actually support plain .gz archives

vhdirk:
I'm abandoning this in favour of https://github.com/containerbase/base/pull/3107

Issue #2078, opened 2024-01-30
Support readonly root filesystem


viceice:
### general

- `/usr/local/bin` is symlinked to `/opt/containerbase/bin`
- `/usr/local/lib` is symlinked to `/opt/containerbase/lib`
- tool prepare state is stored at `/var/lib/containerbase`
- all cache files are stored at `/tmp/containerbase`
- files and folders in user home are symlinked to `/tmp/containerbase/cache` (eg `.cache`, `.npmrc`, `.npm`)

### `binarySource=global`
- needs `/tmp` writable

### `binarySource=install`
- needs `/tmp` writable
- needs `/opt/containerbase` writable
---
*related issues*
- #101
- #380
- #1384
- #1385

rarkins:
It cannot be stopped??

rarkins:
I'm not sure I fully follow, but I guess you mean that when the user mounts a volume to /opt/containerbase then anything we've put there prior is lost. So you're planning to put it in another location and symlink it over whenever the container runs?

viceice:
Yes, that's the core idea.

When later running `install-tool` as user it will simply install to `/opt/containerbase` as now.
I can propably also check and run the `containerbase-cli restore` command implicit from `install-tool`  when something is missing inside `/opt/containerbase`.

rarkins:
Would it be a viable alternative to allow the user to configure the containerbase rw location via env variable? If configured then the restore happens from /opt/containerbase otherwise /opt/containerbase is used? Or is it too hard to set up PATH etc?

viceice:
we can't ensure path then. a user would need to explicit set it 😞

PR #2722, opened 2024-05-30
feat!: redirect home

viceice:
Set user home to `/opt/containerbase/home`

---
- #2078

⚠️ Additional commit message:
```
BREAKING CHANGE: The path `/home/<username>` is now a symlink to `/opt/containerbase/home`.
```


viceice:
won't do

PR #2749, opened 2024-06-04
feat(cli): convert java installer

viceice:
Technically breaking, since i changed the datasource. Should we continue to use `api.adoptium.net` instead?


src/cli/tools/java/index.ts
```diff
@@ -0,0 +1,182 @@
...
+  EnvService,
+  HttpService,
+  PathService,
+} from '../../services';
+import { logger } from '../../utils';
+import type { JavaReleases } from './schema';
+
+const baseUrl =
+  'https://github.com/joschi/java-metadata/raw/main/docs/metadata/ga/linux';
+const fileUrl = 'hotspot/temurin.json';
```
viceice, line 19: @rarkins this is new new source. Should we continue to use `api.adoptium.net` instead?

rarkins: Does it have the same versioning syntax as adoptium? i.e. so that every "install-tool x.y.z+foo" in Dockerfiles today will continue to work?

And will the list of versions be consistent with what adoptium returns? We probably should update Renovate to match sources so that we reduce the chance of a future mismatch

viceice: yes, this file is created from the adoptium api

Issue #1384, opened 2023-08-29
Support install-tool without associated version or env variable


rarkins:
I would like for `install-tool node` to work even if there's no `NODE_VERSION` defined. In such cases Containerbase should resolve the latest stable version and install that. This should work for all `install-tool` commands.

---

- [x] `java` 9fabcd033f4ea79824c56fca94b0bf53a9313aa6
  - [x] `gradle` 9fabcd033f4ea79824c56fca94b0bf53a9313aa6
  - [x] `maven` #2871
- [x] `node` #1828
  - [x] `npm`, `pnpm`, `yarn` ...
- [ ] `php`
  - [x] `composer` #2848
- [ ] `python`
  - [x] `install-pip` #3015
  - [x] `checkov` #3015
  - [x] `copier` #3015
  - [x] `hashin` #3015
  - [x] `pdm` #3015
  - [x] `pip-tools` #3015
  - [x] `pipenv` #3015 
  - [x] `poetry` #3015

viceice:
Should we install lts if available or latest version (eg node v20 vs v21)?

Issue #1074, opened 2023-06-16
Convert tools to ts


viceice:
We can now convert more tools to the new installer

- #771

Issue #815, opened 2023-03-21
Create more prebuild tools


viceice:
We should create more repos to mirror prebuild tools.
Reason: Recent issues with nodejs cdn.

**Tools** (TBF)
- `node`

PR #2848, opened 2024-06-26
feat(php): convert composer installer

viceice:
Use our prebuild composer from https://github.com/containerbase/composer-prebuild

- #815
- #1074
- #1384


viceice:
Need to add a version resolver to support existing `latest` as version

PR #3016, opened 2024-07-30
feat(golang): pick patch version automatically

phil9909:
Currently, it is required to specify the full go version `install-tool golang 1.22.5`.

With this change, the patch version can be left out:  `install-tool golang 1.22` and the latest available patch version will be installed.

A test was added to ensure "1.20" will resolve to "1.20.14". As "1.20" is end of life, there should be no further patches and therefore the test should not break. I guess there is a small chance that the go team decides to patch "1.20", if a serious vulnerability is discovered. It's unlikely, as the EOL date is already several months ago.

### DIFF
--- a/src/usr/local/containerbase/tools/v2/golang.sh
+++ b/src/usr/local/containerbase/tools/v2/golang.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+function check_tool_requirements () {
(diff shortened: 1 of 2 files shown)


viceice:
@phil9909 Please open a new discussion and write your full use-case. I'm closing this for now until we've a conclusion that we like to have this feature

Issue #3277, opened 2024-09-19
Default to high user ID


rarkins:
Some security guides/scanners recommend running containers with a high user ID.

Refs:
- https://docs.prismacloud.io/en/enterprise-edition/policy-reference/kubernetes-policies/kubernetes-policy-index/bc-k8s-37
- https://www.stream.security/rules/ensure-containers-run-with-a-high-uid-to-avoid-host-conflict

Particularly, user 1000 is one of the most likely to have a conflict, but I think we can pick a _really_ high one semi-randomly so that the chance a user needs to use it for something else already is close to zero.

viceice:
do you have a preference?

viceice:
i'll pick `12021` because it's higer than `10000` and this repo was created at `2021` 🙃

PR #3357, opened 2024-10-01
fix: set perms for extracted files

viceice:
related to
- #3277


src/cli/tools/dotnet.ts
```diff
@@ -116,23 +116,9 @@ export class DotnetInstallService extends BaseInstallService {
       file,
       cwd: toolPath,
       strip: 1,
+      options: ['--uid', `${this.envSvc.userId}`, '--gid', '0'],
```
viceice, line 119: @rarkins i wonder if we should set this as default for extracts, so files are always owned by our default user 🤔

PR #3783, opened 2024-12-11
feat: add support for json log formats

nabeelsaabna:
Set log format JSON only when: `CONTAINERBASE_LOG_FORMAT` or the falback `LOG_FORMAT` is set to `json` 
Note: case-insensitive 

image

### DIFF
--- a/src/cli/utils/logger.spec.ts
+++ b/src/cli/utils/logger.spec.ts
@@ -32,4 +32,39 @@ describe('cli/utils/logger', () => {
       },
     );
   });
+
+  test('works - stdout with json', async () => {
+    env.CONTAINERBASE_LOG_FORMAT = 'json';
+    const { pino } = await import('pino');
+    const mod = await import('./logger');
+    expect(mod.logger).toBeDefined();
+    expect(pino).toHaveBeenCalledWith(
+      { level: 'info' },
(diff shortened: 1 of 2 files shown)


viceice (changes requested):
please add a test like
https://github.com/containerbase/base/blob/3c7a21fd6acbeff07617f8de6878add21880b2c5/src/cli/utils/logger.spec.ts#L15-L34

src/cli/utils/logger.spec.ts
```diff
@@ -32,4 +32,39 @@ describe('cli/utils/logger', () => {
       },
     );
   });
+
+  test('works - stdout with json', async () => {
+    env.CONTAINERBASE_LOG_FORMAT = 'json';
```
viceice, line 37: please delete these new variables like the others in `beforeEach`

viceice (commented):
like here

https://github.com/containerbase/base/blob/c23e372f4f0f5e900127e89f82092fc2d54211c9/src/cli/utils/logger.spec.ts#L9-L12

src/cli/utils/logger.spec.ts
nabeelsaabna: done

Issue #2251, opened 2024-02-28
`rust`: use channel file for download infos


viceice:
- https://static.rust-lang.org/dist/channel-rust-1.74.1.toml
- we can cache this file
- needs installer conversion

PR #3785, opened 2024-12-11
feat(rust): support beta and nightly

### LINKED ISSUE #3752: `rust`: support nightly channel
for renovate we need to support nightly channel

- #2251
- https://github.com/renovatebot/renovate/discussions/24870

---

## steps:
- override `check_tool_requirements` to support `nightly` as `TOOL_VERSION`
  https://github.com/containerbase/base/blob/8dc62a6c4af04e012217570404b4d82a6f7c6a98/src/usr/local/containerbase/tools/v2/git-lfs.sh#L3-L5

- Add an install test
   https://github.com/containerbase/base/blob/8dc62a6c4af04e012217570404b4d82a6f7c6a98/test/rust/Dockerfile#L49-L50

viceice:
- Closes #3752

⚠️ Once installed those are currently not updated. So needs short lived containers or use docker sidecar.

### DIFF
--- a/src/usr/local/containerbase/tools/v2/rust.sh
+++ b/src/usr/local/containerbase/tools/v2/rust.sh
@@ -25,6 +25,15 @@ function init_tool () {
   chown -R "${USER_ID}" "${cargo_home}"
 }
 
+function check_tool_requirements () {
+  if [[ "${TOOL_VERSION}" == "nightly" || "${TOOL_VERSION}" == "beta" ]]; then
+    # allow beta and nightly versions
+    return
+  fi
+  # Sensitive default that can be overwritten by tools if needed
+  check_semver "$TOOL_VERSION" "all"
+}
+
 function install_tool () {
   local versioned_tool_path
   local file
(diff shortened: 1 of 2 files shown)


rarkins:
@viceice re:

> Once installed those are currently not updated.

Do you mean by this what once "nightly" is installed, it won't ever be refreshed/updated? Is it something you can address later? Either like:
- Use a static timeout like 24 hours for refreshes, or
- Does `nightly` alias to e.g. `nightly-20241211` and we can cache/bust on that instead of `nightly`?

viceice:
> @viceice re:
> 
> > Once installed those are currently not updated.
> 
> Do you mean by this what once "nightly" is installed, it won't ever be refreshed/updated? Is it something you can address later? Either like:
> 
> * Use a static timeout like 24 hours for refreshes, or
> * Does `nightly` alias to e.g. `nightly-20241211` and we can cache/bust on that instead of `nightly`?

I'll address later. see 
- https://github.com/containerbase/base/issues/2251

PR #3970, opened 2025-01-23
feat(cli): add cleanup paths command

### LINKED ISSUE #3956: Add environment variable for /tmp cleanup at startup
We need to add an opt-in functionality where `/tmp` will be erased at startup/restart each time. Naturally it needs to still work with Containerbase's functionality which _populates_ /tmp at startup too.

Proposed variable name: CONTAINERBASE_TMP_CLEANUP=true

viceice:
- Closes #3956


nabeelsaabna (commented):
Approved

PR #4118, opened 2025-02-14
fix(node): use `NODE_OPTIONS` for `--use-openssl-ca`

### LINKED ISSUE #4033: move `use-openssl-ca` to `NODE_OPTIONS`
- https://github.com/renovatebot/renovate/discussions/33935

viceice:
Pass  `--use-openssl-ca` via `NODE_OPTIONS`, which can be overridden from outer env.

I'm not sure, but `NODE_OPTIONS` can maybe no longer set via renovate options. 🤔 

- closes #4033

### DIFF
--- a/src/cli/tools/node/index.ts
+++ b/src/cli/tools/node/index.ts
@@ -36,6 +36,7 @@ export class NodePrepareService extends BasePrepareService {
         NO_UPDATE_NOTIFIER: '1',
         npm_config_update_notifier: 'false',
         npm_config_fund: 'false',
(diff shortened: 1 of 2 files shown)


rarkins (approved):

rarkins:
How confident can we be that there's no regression?

viceice:
> How confident can we be that there's no regression?

🤷‍♂️ I think it can only break people which already use `NODE_OPTIONS` and depend on `use-openssl-ca` but don't pass them via `NODE_OPTIONS`

PR #5434, opened 2025-11-20
feat(docker): add buildx and docker-compose

### LINKED ISSUE #3603: Add Docker buildx plugin when installing Docker
We have some post-upgrade commands that execute `docker build` to rebuild a base image before executing some commands, and those Dockerfiles have `RUN --mount` instructions that requires buildx to be installed. Could the docker buildx plugin be added as part of the `docker` tool here: https://github.com/containerbase/base/blob/main/src/cli/tools/docker.ts?

viceice:
- https://github.com/containerbase/devcontainer/issues/976
- closes #3603

### DIFF
--- a/src/cli/tools/docker/compose.ts
+++ b/src/cli/tools/docker/compose.ts
@@ -21,7 +21,7 @@ export class DockerComposeInstallService extends BaseInstallService {
   }
 
   override async install(version: string): Promise<void> {
-    const baseUrl = `https://github.com/docker/${this.name}/releases/download/v${version}/`;
+    const baseUrl = `https://github.com/docker/compose/releases/download/v${version}/`;
     const filename = `${this.name}-linux-${this.ghArch}`;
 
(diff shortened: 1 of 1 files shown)


src/cli/tools/docker/buildx.ts
```diff
@@ -0,0 +1,69 @@
...
+@injectable()
+@injectFromHierarchy()
+export class BuildxInstallService extends BaseInstallService {
+  readonly name = 'buildx';
+
+  override readonly parent = 'docker';
+
+  override async install(version: string): Promise<void> {
+    const baseUrl = `https://github.com/docker/${this.name}/releases/download/v${version}/`;
+    const filename = `${this.name}-v${version}.linux.${this.envSvc.arch}`;
```
astellingwerf:
```suggestion
    const filename = `${this.name}-v${version}.linux-${this.envSvc.arch}`;
```

PR #6071, opened 2026-03-02
test(tools/node): document existing behaviour for `NODE_OPTIONS` in `env.sh`

jamietanna:
As found in #5962.


jamietanna:
Oops, expected this to need a review 🙇

PR #6569, opened 2026-04-29
feat(haskel): support cabal v1 and v2

### LINKED ISSUE #5808: Add support for new tool Haskell
Haskell (GHCup installer) and its package manager `Cabal`

* support `install-tool`
* support `uninstall-tool`

viceice:
cabal v1 and v2 need an older ghc (compiler) which isn't available 

- closes #5808

### DIFF
--- a/test/haskell/Dockerfile
+++ b/test/haskell/Dockerfile
@@ -43,13 +43,22 @@ FROM base AS build
 RUN prepare-tool cabal ghc
 
 #--------------------------------------
-# test: haskell (root,open-shift)
+# test-ghc
 #--------------------------------------
-FROM base AS haskell-a
+FROM build AS test-ghc
 
-RUN env
 # renovate: datasource=docker packageName=haskell
 RUN install-tool ghc 9.14.1
+
+RUN set -ex; \
+  ghc --version; \
+  ghc-pkg dot --global; \
+  true;
+
+#--------------------------------------
+# test: haskell (root,open-shift)
+#--------------------------------------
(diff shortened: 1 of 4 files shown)


test/haskell/Dockerfile
```diff
@@ -97,10 +106,49 @@ RUN set -ex; \
...
+  cabal sandbox init; \
+  true;
+#--------------------------------------
+# test: cabal v2 (root,open-shift)
+#--------------------------------------
+FROM test-ghc AS cabal-v2
+
+# cabal v2 need ghc <8.7
+# no binaries yet
+# RUN install-tool ghc 8.8.4
```
jamietanna, line 131: Just to confirm, does something like https://www.haskell.org/ghc/download_ghc_8_4_4.html#linux_x86_64 as a pre-built binary not work for this? I assume not, but wanted to check

I guess potentially not as it's Debian, not Ubuntu based, so system libraries may have drifted - especially by now

viceice: yes, it doesn't incluse the bin folder like the v9 builds :-/

at least i have no idea how to get the `ghc` and `ghc-pkg` binaries whiout a full rebuild.

do you have a more haskell experienced user at hand to help?

viceice: at least google AI tells me that we need make and configure

```sh
./configure
sudo make install
```

😕

jamietanna: Let me check

PR #6768, opened 2026-05-22
fix(rust): add support for `nightly` channel

jamietanna:
As per https://github.com/renovatebot/renovate/discussions/43489, when
Renovate provides a date-based nightly version, such as
`nightly-2026-06-19`, Containerbase doesn't currently determine the
correct URL to download the nightly build from.

We can make sure to rewrite the URL as appropriate when a specific
Nightly version is specified.

With help from Claude to write the test, and GPT-4.1 to write the
pattern match.

### DIFF
--- a/test/rust/Dockerfile
+++ b/test/rust/Dockerfile
@@ -117,6 +117,19 @@ RUN rustc --version
 RUN cargo --version
 
 #--------------------------------------
+# test e: nightly with date
+#--------------------------------------
+FROM base AS teste
+
+USER 12021
+
+RUN install-tool rust nightly-2026-06-19
+
+SHELL [ "/bin/sh", "-c" ]
+RUN rustc --version
+RUN cargo --version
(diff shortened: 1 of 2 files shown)


test/rust/Dockerfile
```diff
@@ -116,6 +116,19 @@ SHELL [ "/bin/sh", "-c" ]
...
 RUN cargo --version
 
+#--------------------------------------
+# test e: nightly with date
+#--------------------------------------
+FROM base AS teste
+
+USER 12021
+
+RUN install-tool rust nightly-2026-06-19
```
jamietanna, line 126: https://static.rust-lang.org/manifests.txt indicates that these versions are never deleted, so we're probably safe to keep this hardcoded, instead of updating through Renovate (maybe weekly/monthly)?

viceice (approved):

PR #6880, opened 2026-06-12
feat(erlang): make OTP 27 the default for Elixir

kivra-fabber:
This should make https://github.com/containerbase/base/pull/6850 happy

As of 1.20 Elixir does not support OTP 26 anymore so that should not be the default choice here.

https://elixir.hexdocs.pm/1.20.1/compatibility-and-deprecations.html#between-elixir-and-erlang-otp

I see renovate jobs failing because when installing Elixir 1.20 they fail to download elixir-otp-26.zip from the releases. See assets here for example: https://github.com/elixir-lang/elixir/releases/tag/v1.20-latest

### DIFF
--- a/src/usr/local/containerbase/tools/v2/elixir.sh
+++ b/src/usr/local/containerbase/tools/v2/elixir.sh
@@ -35,7 +35,7 @@ function install_tool () {
   local versioned_tool_path
   local file
   local base_url="https://github.com/elixir-lang/elixir/releases/download"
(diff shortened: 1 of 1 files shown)


viceice (approved):

kivra-fabber:
I see this cannot be merged without verified commits.

Opened up a new PR with no unverified commits: https://github.com/containerbase/base/pull/6924

kivra-fabber:
Superseded by #6924

PR #7278, opened 2026-08-10
feat: add support for GitHub CLI `gh`

### LINKED ISSUE #7154: New tool: GitHub CLI (gh)
(Co-authored-by: Claude Opus 4.8)

### Tool name

GitHub CLI (`gh`)

### What is the tool used for?

Interacting programmatically with the GitHub APIs

### Tool scope

- [x] Runtime only (CLI / binary)
- [ ] Full development stack (headers, SDK, compiler)

`gh` ships as a single, statically-linked Go binary — no SDK, headers, or
compiler are required.

### Supported version range

All available `v2.x` releases.

Preferably the latest `v2` major line (`>=2.0.0 <3.0.0`)

### Key commands that must work

```sh
gh --version
gh auth status
gh api /user
gh extension upgrade aw --dry-run
```

### Uninstall strategy

`gh` is self-contained. Following the containerbase versioned-tool layout (as
used by e.g. `bazelisk`), removal is just deleting the versioned tool path and
the shell wrapper:

```sh
rm -rf /opt/containerbase/tools/gh
rm -f /usr/local/bin/gh
```

### License of the tool (and any pre-built binaries)

MIT — https://github.com/cli/cli/blob/trunk/LICENSE

### Minimal Dockerfile snippet (optional but highly encouraged)

`gh` publishes pre-built `linux` tarballs (amd64/arm64) on its GitHub releases
page, so no third-party repo or package manager is needed:

```sh
FROM ubuntu:24.04 AS base
ARG GH_VERSION=2.62.0
ARG TARGETARCH=amd64
RUN set -eux; \
    curl -fsSL "https://github.com/cli/cli/releases/download/v${GH_VERSION}/gh_${GH_VERSION}_linux_${TARGETARCH}.tar.gz" \
      -o /tmp/gh.tar.gz; \
    tar -xzf /tmp/gh.tar.gz -C /tmp; \
    install -m 0755 "/tmp/gh_${GH_VERSION}_linux_${TARGETARCH}/bin/gh" /usr/local/bin/gh; \
    rm -rf /tmp/gh.tar.gz "/tmp/gh_${GH_VERSION}_linux_${TARGETARCH}"; \
    gh --version
```

This closely mirrors the existing `bazelisk` installer (single binary pulled
from `github.com/<org>/<repo>/releases`).

### Suggested tests

```sh
# Test 1 – Runtime availability / correct version resolved
RUN gh --version

# Test 2 – Binary is functional and can reach the API layer
#          (unauthenticated call still exercises the HTTP client + TLS)
RUN gh api rate_limit --jq '.resources.core.limit'
```

### Any other relevant information

- Release assets are provided per-arch (`linux_amd64`, `linux_arm64`), matching
  containerbase's multi-arch build matrix.
- Archives contain man pages and shell completions under `share/`; these can be
  dropped to keep the image slim, only `bin/gh` is required at runtime.
- `gh` reads credentials from the `GH_TOKEN` / `GITHUB_TOKEN` environment
  variables, so no interactive `gh auth login` is needed in CI/automation.
- Implementation should be close to the existing `bazelisk` tool
  (`src/cli/tools/bazelisk.ts`): download from GitHub releases, place under a
  versioned tool path, and expose via a shell wrapper.

jamietanna:
As per #7154, we can add support for the GitHub CLI, so it can be used
for other purposes.

We're only supporting the v2 of the GitHub CLI, and don't currently
expose any means for installing extensions.

Closes #7154.


test/latest/Dockerfile
```diff
@@ -291,6 +294,8 @@ USER 12021
 
 RUN bazel --version
 
+RUN gh --version
```
viceice, line 299: what about install a extension and check version of it too?

viceice (commented):
otherwise LGTM

test/latest/Dockerfile
jamietanna: Install an extension at a specific version? Or just whatever is latest?

.github/renovate.json
```diff
@@ -2,7 +2,8 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "github>containerbase/.github",
-    "github>containerbase/.github//merge-queue.json"
+    "github>containerbase/.github//merge-queue.json",
+    "customManagers:dockerfileVersions"
```
viceice: I don't think we need this, there should be a similar custom manager in our base preset

viceice (approved):

PR #7338, opened 2026-09-02
feat(tools): install Vite+ releases

afonsojramos:
## Summary

Containerbase can install an exact Vite+ release as `vp`, giving downstream tools a verified way to run the compatibility planner owned by that release.

This is the second part of the Vite+ dependency-alignment integration. The planner from voidzero-dev/vite-plus#2600 is now available in Vite+ 0.3.1. The downstream Renovate integration is in renovatebot/renovate#45630.

## Design

- Installation selects the official archive for the current Linux architecture and validates it against `vp-checksums.txt`.
- The extracted `vp`, planner sidecar, and toolchain manifest retain the release layout expected by the global CLI.
- Vite+ declares Node as its parent tool so the JavaScript planner always has a compatible runtime.
- Releases that predate the planner contract return a stable marker. Renovate can treat that case as unsupported instead of as a failed update.

## Validation

- All 141 Vitest tests passed, including the 8 Vite+ installer tests.
- Formatting, ESLint, both TypeScript configurations, Markdown lint, and `git diff --check` passed.
- The CLI build passed. Both Node Docker test images (`linux/amd64` and `linux/arm64`) installed Vite+ 0.3.1 and ran the bundled `sync-versions --json` planner as a non-root user.
- The planner smoke test checks dependency alignment, repeat execution, and unchanged project files. A deliberately wrong expected version failed as expected.

## Summary by CodeRabbit

- **New Features**
  - Added support for installing, linking, and testing the Vite+ (`vp`) tool.
  - Added architecture-specific downloads with checksum verification.
  - Added version synchronization for Vite+, Vite, Vitest, and coverage tooling.

- **Documentation**
  - Documented Vite+ release archives and checksum locations for supported Linux architectures.

- **Tests**
  - Added integration and container coverage for Vite+ installation and version synchronization.
  - Added validation for supported architectures, checksums, downloads, linking, and version checks.

### DIFF
--- a/docs/custom-registries.md
+++ b/docs/custom-registries.md
@@ -776,6 +776,20 @@ Samples:
 https://github.com/vmware-tanzu/carvel-vendir/releases/download/v0.22.0/vendir-linux-amd64
 ```
 
+## `vp`
+
+Vite+ releases are downloaded from:
+
+- `https://github.com/voidzero-dev/vite-plus/releases`
+
+Release archives and their checksum manifest follow these paths:
+
+```txt
+https://github.com/voidzero-dev/vite-plus/releases/download/v<version>/vp-x86_64-unknown-linux-gnu.tar.gz
+https://github.com/voidzero-dev/vite-plus/releases/download/v<version>/vp-aarch64-unknown-linux-gnu.tar.gz
+https://github.com/voidzero-dev/vite-plus/releases/download/v<version>/vp-checksums.txt
+```
+
 ## `wally`
 
 Wally releases are downloaded from:
--- a/test/Dockerfile.distro
+++ b/test/Dockerfile.distro
@@ -211,6 +211,13 @@ RUN install-tool pnpm 10.34.5
 # renovate: datasource=npm packageName=@yarnpkg/cli-dist
 RUN install-tool yarn 4.18.0
 
+# renovate: datasource=github-releases depName=vp packageName=voidzero-dev/vite-plus
+ARG VP_VERSION=0.3.1
+RUN install-tool vp "${VP_VERSION}"
+
+COPY test/vp/sync-versions.mjs /test/vp-sync-versions.mjs
+RUN node /test/vp-sync-versions.mjs "${VP_VERSION}"
+
 #--------------------------------------
 # Image: test-php
 #--------------------------------------
--- a/test/latest/Dockerfile
+++ b/test/latest/Dockerfile
@@ -387,6 +387,25 @@ RUN install-tool nuget 7.6.0
 RUN nuget search nuget.commandline -Take 1
 
 #--------------------------------------
+# Image: vp
+#--------------------------------------
+FROM base AS test-vp
+
+# renovate: datasource=github-releases packageName=containerbase/node-prebuild versioning=node
+RUN install-tool node 24.19.0
+
+# renovate: datasource=github-releases depName=vp packageName=voidzero-dev/vite-plus
+ARG VP_VERSION=0.3.1
+RUN install-tool vp "${VP_VERSION}"
+
+COPY test/vp/sync-versions.mjs /test/vp-sync-versions.mjs
+
+USER 12021
+SHELL ["/bin/sh", "-c"]
+
+RUN node /test/vp-sync-versions.mjs "${VP_VERSION}"
+
+#--------------------------------------
 # final
 #--------------------------------------
 FROM base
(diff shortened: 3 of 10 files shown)


afonsojramos:
Upstream update: voidzero-dev/vite-plus#2600 has merged. This draft is now waiting for a Vite+ release containing `vp sync-versions --json` and the corresponding release artifacts.

Once that ships, I'll validate `install-tool vp <version>` against the published archive and checksums before marking this ready. The Renovate integration is in renovatebot/renovate#45630.

docs/custom-registries.md
```diff
@@ -776,6 +776,20 @@ Samples:
...
+Vite+ releases are downloaded from:
+
+- `https://github.com/voidzero-dev/vite-plus/releases`
+
+Release archives and their checksum manifest follow these paths:
+
+```txt
+https://github.com/voidzero-dev/vite-plus/releases/download/v<version>/vp-x86_64-unknown-linux-gnu.tar.gz
+https://github.com/voidzero-dev/vite-plus/releases/download/v<version>/vp-aarch64-unknown-linux-gnu.tar.gz
+https://github.com/voidzero-dev/vite-plus/releases/download/v<version>/vp-checksums.txt
```
viceice: please use real version instead of placeholder like the other samples

test/latest/Dockerfile
viceice, line 1: move to node, as this depends on node

src/cli/tools/node/vp.ts
viceice, line 1: move to node folder, as the parent tool is node

test/Dockerfile.distro
```diff
@@ -211,6 +211,13 @@ RUN install-tool pnpm 10.34.5
 # renovate: datasource=npm packageName=@yarnpkg/cli-dist
 RUN install-tool yarn 4.18.0
 
+# renovate: datasource=github-releases depName=vp packageName=voidzero-dev/vite-plus
+ARG VP_VERSION=0.3.1
+RUN install-tool vp "${VP_VERSION}"
+
+COPY test/vp/sync-versions.mjs /test/vp-sync-versions.mjs
+RUN node /test/vp-sync-versions.mjs "${VP_VERSION}"
+
```
viceice: no real tests here

afonsojramos: Just to clarify, did you mean functional tests shouldn’t run in Dockerfile.distro, or were you questioning the value of the test itself?

I’ve moved it to the Node images and kept the distro checks installation-only. The smoke test verifies that the installed release can run its bundled planner as a non-root user, which vp --version alone wouldn’t cover.

afonsojramos:
@viceice, can you take another look? Thanks :pray:

test/Dockerfile.distro
viceice: yes, no functional tests on distro. it's only to check if the tool can generally be installed on other ubuntu versions.

test/node/vp/sync-versions.mjs
```diff
@@ -0,0 +1,86 @@
...
+    '@vitest/coverage-v8': '0.0.0',
+    typescript: '5.8.0',
+  },
+};
+const contents = JSON.stringify(manifest, null, 2) + '\n';
+const config = 'throw new Error("Project configuration must not be loaded");\n';
+
+function plan(manifestContents) {
+  return JSON.parse(
+    execFileSync(process.env.VP_TEST_BIN ?? 'vp', ['sync-versions', '--json'], {
```
viceice: use async io (promises) and execa

test/Dockerfile.distro
```diff
@@ -211,6 +211,10 @@ RUN install-tool pnpm 10.34.5
 # renovate: datasource=npm packageName=@yarnpkg/cli-dist
 RUN install-tool yarn 4.18.0
 
+# renovate: datasource=github-releases depName=vp packageName=voidzero-dev/vite-plus
+ARG VP_VERSION=0.3.1
+RUN install-tool vp "${VP_VERSION}"
```
viceice:
```suggestion
# renovate: datasource=github-releases packageName=voidzero-dev/vite-plus
RUN install-tool vp 0.3.1
```
use simple variant

afonsojramos:
@viceice Thanks for the review and clarification! The distro checks are installation-only and now use the simpler versioned install you suggested. The functional checks remain in the Node images.

package.json
```diff
@@ -13,6 +13,7 @@
     "bats": "node tools/bats.js --timing --verbose-run",
     "build": "run-s 'build:*'",
     "build:cli": "node tools/build.js",
+    "build:test": "rolldown test/node/vp/sync-versions.mjs --platform node --format esm --file dist/test/vp-sync-versions.mjs",
```
viceice: why rolldown? our node version should be new enough to allow esm with top-level await by default

afonsojramos: Apologies, had this locally for testing. Have undone the change

PR #7410, opened 2026-09-15
feat: add `list tools` command

### LINKED ISSUE #6429: Add a `tools list` command
To provide a way for users to see what versions of tools are installed on their Containerbase image, we should provide a `containerbase-cli list tools` command (and potentially a `list-tools` command, too)

Additionally, as a way to provide programmatic access to the tools themselves, we should have a `--json` flag.

The output for the JSON format should be documented, ideally in a JSON schema, but that can be a follow-up.

Given Containerbase logs on `stdout`, we should have a `--out` flag to output the JSON to a file path.

This will allow users to understand what's on the image, without needing to delve into the `.nedb` files in `/opt/containerbase/data`.

viceice:
Closes #6429

Adds `containerbase-cli list tools`, so users can see which tools and versions are installed on an image without having to inspect the `.nedb` files in `/opt/containerbase/data`.

## Usage

```bash
$ containerbase-cli list tools
NAME      VERSION  OTHER VERSIONS
java-jdk  -        21.0.12+7
node      22.11.0  20.11.0
pnpm      10.0.1
```

The table shows the currently linked version and any other installed versions. A `-` means no version of the tool is currently linked, which happens when the linked version is uninstalled while other versions remain installed, as nothing relinks one of them. Tools and versions are sorted alphabetically with numbers in numeric order, so `9.0.0` comes before `10.0.0`.

`--json` emits the list for programmatic use:

```bash
$ containerbase-cli list tools --json
{
  "tools": [
    {
      "name": "java-jdk",
      "version": null,
      "versions": [{ "version": "21.0.12+7" }]
    },
    {
      "name": "node",
      "version": "22.11.0",
      "versions": [{ "version": "20.11.0" }, { "version": "22.11.0" }]
    },
    {
      "name": "pnpm",
      "version": "10.0.1",
      "versions": [
        {
          "version": "10.0.1",
          "parent": { "name": "node", "version": "22.11.0" }
        }
      ],
      "type": "npm"
    }
  ]
}
```

Since containerbase logs to `stdout`, `--out <file>` writes the output to a file instead, minified:

```bash
containerbase-cli list tools --json --out /tmp/tools.json
```

## Implementation

- `VersionService.listInstalled()` joins the `versions`, `state` and `types` databases into a single list. The linked version is resolved by `tool.name`, so tools linked under an alias — `java-jdk` is linked as `java` — report their version instead of `-`.
- `ListToolsCommand` renders that list as a table or JSON and writes it to stdout or to `--out`. Status logs are suppressed under `--json`, since containerbase logs to stdout and would corrupt the output.
- `src/cli/services/version.schema.ts` holds the zod schemas as the single source of truth, with the TypeScript types inferred from them.

## JSON schema

The output is documented by `docs/list-tools.schema.json`, generated from the `InstalledTools` zod schema by `pnpm schema` (`tools/schema.ts`, which formats with prettier so the file lands ready to commit).

`pnpm lint:schema` regenerates it and fails when the result differs from what is committed; it runs in the CI lint job, so the committed schema can't go stale.

## Not included

A `list-tools` alias binary. The issue mentions it as a possibility; left out, it can be added later by registering a CLI mode and a symlink in `install-containerbase.sh`.

## Test plan

- `src/cli/command/list-tools.spec.ts` covers the table output, JSON output, `--out` (including that the file is minified), the empty listing, alias resolution, a non-error failure, schema conformance of the output, and that the committed schema is up to date. `list-tools.ts`, `version.schema.ts` and `version.service.ts` are at 100% coverage.
- The `test/node` container test asserts the listing before `uninstall-tool node -a -r` and `No tools installed.` afterwards. Verified locally with `pnpm test:docker -b -t test-x86_64 node`.
- `pnpm vitest run` — 38 files, 144 tests passing.
- `pnpm tsc --noEmit`, `pnpm eslint`, `pnpm prettier`, `pnpm lint:markdown`, `pnpm lint:schema` — clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/cli/command/list-tools.ts
```diff
@@ -0,0 +1,96 @@
...
+          `Listing tools failed in ${prettyMilliseconds(Date.now() - start)}.`,
+        );
+      } else {
+        logger.debug(
+          `Listing tools succeeded in ${prettyMilliseconds(Date.now() - start)}.`,
+        );
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
viceice: changing logging to stderr would be a breaking charge and needs to be deferred to a later major release.

in meantime users should use the out parameter.

jamietanna (commented):
I'd like us to have a JSON Schema for the JSON output, ideally, so we have some strong typing + documentation we can use with it

docs/tools.md
```diff
@@ -370,3 +372,59 @@ Uninstalls a pip package from the container.
...
+
+`$ containerbase-cli list tools [-j,--json] [-o,--out <file>]`
+
+#### Examples <a name="Examples-list-tools"></a>
+
+- Lists all installed tools
+
+  ```bash
+  $ containerbase-cli list tools
+  node  22.11.0 (20.11.0)
```
jamietanna: I'd say we should clarify i.e.:

```suggestion
  node  22.11.0 (Other installed versions: 20.11.0)
```

src/cli/command/list-tools.ts
```diff
@@ -0,0 +1,96 @@
...
+      ['List all installed tools', '$0 list tools'],
+      ['List all installed tools as json', '$0 list tools --json'],
+      [
+        'Write the json output to a file',
+        '$0 list tools --json --out /tmp/tools.json',
+      ],
+    ],
+  });
+
+  json = Option.Boolean('-j,--json', false, {
```
jamietanna: IMO we don't need a short `-j` flag now - let's leave it as a long flag for now

src/cli/command/list-tools.ts
```diff
@@ -0,0 +1,96 @@
...
+    description: 'Outputs the tool list as json.',
+  });
+
+  out = Option.String('-o,--out', {
```
jamietanna: As above, may not be worth adding a short flag for now

jamietanna (commented):
We should probably also reference `parent`/`parents` for a tool that has a paren?

src/cli/command/list-tools.ts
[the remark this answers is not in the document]
viceice: yes, if non exist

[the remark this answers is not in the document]
viceice: update the new issue so that general logging will go to stderr instead of stdout. not only for list tools command

docs/list-tools.schema.json
```diff
@@ -0,0 +1,63 @@
...
+      "items": {
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string",
+            "description": "tool name"
+          },
+          "version": {
+            "description": "the currently linked version, null if the tool isn't linked",
+            "type": ["string", "null"]
```
jamietanna, line 16: When would this be `null`?

src/cli/command/list-tools.ts
```diff
@@ -0,0 +1,102 @@
...
+      const container = createContainer();
+      const versionSvc = await container.getAsync(VersionService);
+      const tools = await versionSvc.listInstalled();
+
+      const output = this.json
+        ? `${JSON.stringify({ tools }, null, 2)}\n`
+        : toText(tools);
+
+      if (this.out) {
+        await writeFile(this.out, output, { encoding: 'utf8' });
```
jamietanna, line 85: We should write the non-pretty JSON when writing to a file

docs/list-tools.schema.json
viceice: When the linked version is uninstalled while other versions of the tool remain installed.

`InstallToolService.uninstall()` calls `removeCurrent()` when the version being removed is the current one, but nothing re-links one of the remaining versions. So:

```bash
install-tool node 20.11.0
install-tool node 22.11.0     # current = 22.11.0
uninstall-tool node 22.11.0   # state row dropped, 20.11.0 still installed
```

leaves `node` in the `versions` database with no `state` row, and `list tools` reports it as `node  - (Other installed versions: 20.11.0)` — installed, but nothing on `PATH` for it.

There was also an unintended way to hit this, which your question turned up: java tools are linked under their `java` alias, so the state row is `{ name: 'java', tool: { name: 'java-jdk', ... } }`, and the lookup by state name missed it — `java-jdk` was reported as unlinked even when it was on `PATH`. Fixed in 5ce842e by resolving the current version via `tool.name`, with a regression test.

<sub>Comment written by Claude Code on behalf of @viceice.</sub>

viceice: should we make the output more table like? so we've headers? :thinking:

docs/list-tools.schema.json
```diff
@@ -0,0 +1,63 @@
...
+                  },
+                  "required": ["name", "version"],
+                  "additionalProperties": false,
+                  "description": "the tool this version was installed for, eg. the node version a npm package was installed with"
+                }
+              },
+              "required": ["version"],
+              "additionalProperties": false
+            },
+            "description": "all installed versions, sorted ascending"
```
jamietanna: Words to this effect:

```suggestion
            "description": "all installed versions, sorted ascending alphabetical order"
```

As far as I'm aware, we're not doing anything with Renovate's versioning knowledge, so it's not _necessarily_ in sorted order per the versioning scheme

viceice: we do  `numerical:true` sorting, so: 1 < 9 < 10 < 11 < 20 < 20.1

docs/list-tools.schema.json
jamietanna: Yes I think that's probably best - if there's no chance that a tool may have a duplicate name? What happens if two ecosystems have the same package name? What's the tool name then?

docs/list-tools.schema.json
jamietanna: OK, might be worth clarifying that

docs/tools.md
```diff
@@ -370,3 +372,66 @@ Uninstalls a pip package from the container.
...
+`$ containerbase-cli list tools [--json] [--out <file>]`
+
+#### Examples <a name="Examples-list-tools"></a>
+
+- Lists all installed tools
+
+  ```bash
+  $ containerbase-cli list tools
+  node  22.11.0 (Other installed versions: 20.11.0)
+  pnpm  10.0.1
```
jamietanna: Let's add an example of what it looks like when it's not linked

docs/list-tools.schema.json
```diff
@@ -0,0 +1,63 @@
...
+                      "type": "string",
+                      "description": "tool version"
+                    }
+                  },
+                  "required": ["name", "version"],
+                  "additionalProperties": false,
+                  "description": "the tool this version was installed for, eg. the node version a npm package was installed with"
+                }
+              },
+              "required": ["version"],
```
jamietanna, line 44: per below, what if it's not linked? Does that mean this isn't then set?

docs/tools.md
```diff
@@ -370,3 +372,66 @@ Uninstalls a pip package from the container.
...
+
+Lists all installed tools and their versions.
+
+The currently linked version is printed first, additional installed versions follow in parentheses.
+A `-` is printed when no version of the tool is currently linked.
+
+With `--json` the list is printed as JSON, so it can be processed programmatically.
+The JSON output is described by [`list-tools.schema.json`](./list-tools.schema.json), which is generated from the `InstalledTools` schema in `src/cli/services/version.schema.ts`.
+Because containerbase logs to `stdout`, use `--out` to write the output to a file instead.
+JSON written to a file isn't pretty printed.
```
jamietanna: Alternatively?

```suggestion
When writing to a file, JSON output is minified.
```

docs/tools.md
```diff
@@ -370,3 +372,66 @@ Uninstalls a pip package from the container.
...
+
+#### Description <a name="Description-list-tools"></a>
+
+Lists all installed tools and their versions.
+
+The currently linked version is printed first, additional installed versions follow in parentheses.
+A `-` is printed when no version of the tool is currently linked.
+
+With `--json` the list is printed as JSON, so it can be processed programmatically.
+The JSON output is described by [`list-tools.schema.json`](./list-tools.schema.json), which is generated from the `InstalledTools` schema in `src/cli/services/version.schema.ts`.
```
jamietanna: Not sure we need all the detail


```suggestion
The JSON output is described by [the `list-tools.schema.json` JSON Schema](./list-tools.schema.json).
```

docs/tools.md
```diff
@@ -370,3 +372,66 @@ Uninstalls a pip package from the container.
...
+  ```bash
+  $ containerbase-cli list tools
+  node  22.11.0 (Other installed versions: 20.11.0)
+  pnpm  10.0.1
+  ```
+
+- Lists all installed tools as JSON
+
+  ```bash
+  $ containerbase-cli list tools --json
```
jamietanna, line 418: as above, let's show what it looks like when one is unlinked

tools/schema.ts
```diff
@@ -0,0 +1,10 @@
+import { writeFile } from 'node:fs/promises';
+import { z } from 'zod';
+import { InstalledTools } from '../src/cli/services/version.schema.ts';
+
+// generates the json schema documenting `containerbase-cli list tools --json`
+await writeFile(
+  new URL('../docs/list-tools.schema.json', import.meta.url),
+  `${JSON.stringify(z.toJSONSchema(InstalledTools), null, 2)}\n`,
```
jamietanna: Nice use of not having to write it ourselves 🥳

jamietanna (changes requested):
We'll need a CI check to make sure that the generated JSON Schema that's committed isn't out-of-sync with what the working tree has

And a few other comments, otherwise looks good

src/cli/command/list-tools.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    test('json output matches the schema', async () => {
+      const stdout = new StdoutMock();
+
+      expect(await cli.run(['list', 'tools', '--json'], { stdout })).toBe(0);
+
+      const output: unknown = JSON.parse(stdout.output);
+      // parsing strips unknown keys, so an equal result means no extra fields
+      expect(InstalledTools.parse(output)).toEqual(output);
+    });
```
viceice, line 136: @jamietanna here are the schema tests

docs/list-tools.schema.json
viceice: duplicate tool names are currently not possible

docs/list-tools.schema.json
viceice: it's null

.github/workflows/build.yml
```diff
@@ -51,6 +51,7 @@ jobs:
           pnpm prettier
           pnpm lint:markdown
           pnpm lint:types
+          pnpm lint:schema
```
jamietanna, line 54: We still need to check there are no uncommitted files - i.e. https://www.jvt.me/posts/2022/04/29/git-uncommitted-changes/

jamietanna (commented):
Otherwise LGTM

.github/workflows/build.yml
viceice: 🤔  isn't this enough: `"lint:schema": "pnpm schema && git diff --exit-code docs/list-tools.schema.json",`?

jamietanna: Bah, missed that we had that, sorry!

jamietanna (approved):

